Issue #5, opened 2026-04-09
Everything32.dll missing from repo - build warns/fails on fresh clone


msarson:
After pulling in the SQLite DLL fix (#4), the next build blocker is \Everything32.dll\ not being present in the repo.

The csproj has:
\\\xml
<Content Include="Everything32.dll">
  <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
</Content>
\\\

This is the VoidTools Everything SDK DLL used by \EverythingService.cs\. It's a runtime dependency (not a compile reference), so the C# code compiles fine, but MSBuild fails at the copy step:

\\\
error MSB3030: Could not copy the file 'Everything32.dll' because it was not found.
\\\

**Workaround applied in my fork:** Made the Content include conditional on the file existing (\Condition="Exists('Everything32.dll')"\) so fresh-clone builds succeed. Everything search won't work at runtime without the DLL, but at least the build doesn't fail.

Could you commit \Everything32.dll\ to the repo (similar to the SQLite DLLs) or document where to obtain it? It should be available from the VoidTools Everything SDK: https://www.voidtools.com/support/everything/sdk/

PR #7, opened 2026-04-09
Add Directory.Build.props for portable ClarionRoot configuration

msarson:
## Problem

Building from a fresh clone requires Clarion to be installed at exactly John's paths (\C:\Clarion12\, \C:\Clarion11-13372\, \C:\Clarion10\). Anyone with a different install location gets immediate compile failures because the Clarion IDE reference assemblies can't be found.

## Solution

Add a \Directory.Build.props\ at the repo root that:
- Provides per-version \ClarionRoot\ defaults matching John's layout (no behaviour change for existing builds)
- Supports a gitignored \Directory.Build.props.user\ override for developers with non-standard paths
- Removes the now-duplicate \ClarionVersion\/\ClarionRoot\ property definitions from \ClarionAssistant.csproj\

The \.user\ file pattern is well-established in MSBuild (e.g. \*.csproj.user\) — it's already gitignored via the \*.user\ rule.

Also makes the \Everything32.dll\ content copy conditional on file existence so the build doesn't hard-fail if the VoidTools SDK DLL is absent (fixed separately in #5, but belt-and-suspenders).

## README

Added a 'Configuring your Clarion path' section to the Building from Source docs explaining the \.user\ file and the command-line override alternative.

## Testing

Built successfully on a machine with Clarion at \C:\Clarion\Clarion11.1\ using \Directory.Build.props.user\ override.

### DIFF
--- a/ClarionAssistant/ClarionAssistant.csproj
+++ b/ClarionAssistant/ClarionAssistant.csproj
@@ -155,7 +155,7 @@
       <Link>docs\ClarionAssistant-Guide.html</Link>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
-    <Content Include="Everything32.dll">
+    <Content Include="Everything32.dll" Condition="Exists('Everything32.dll')">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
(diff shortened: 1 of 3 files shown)


peterparker57 (approved):
Clean PR — follows standard MSBuild conventions, no behavior change for existing builds. The .user file pattern is the right approach for per-developer path overrides. README updates are helpful for contributors.

PR #8, opened 2026-04-10
feat: Safe embeditor read/write tools for AI assistants

msarson:
## Summary

Adds four new MCP tools that give Claude Code a safe, token-efficient way to read and write Clarion procedure embed code directly through the PWEE embeditor, without touching any files on disk.

---

## The problem with the current approach

When Claude needs to edit embed code today, the only option is \get_active_file\ — which dumps the **full generated Clarion source** (typically 40–90 KB of generated boilerplate) into Claude's context window. Claude then has to:

1. Parse thousands of lines of generated code to find the few embed slots it cares about
2. Guess at line numbers that may shift after any edit
3. Use raw \
eplace_range\ / \insert_text_at_cursor\ operations that operate on generated text Claude was never meant to edit directly

This is expensive in tokens, fragile, and risks corrupting generated code.

---

### \get_embeditor_source\
Returns the full PWEE source but with **embed slots annotated** using tokens:
- \«E:N/»\ — an empty (unfilled) embed point at 1-based line N
- \«E:N»\...\«/E:N»\ — a filled embed containing user code

Noise lines (\! Start of generated...\, \! End of...\, \! [Priority N]\, \!!!\) are stripped. This turns a 90 KB wall of generated source into a structured map of exactly where user code lives.

### \search_embeditor_source\
**The primary tool for targeted edits.** Accepts a regex pattern and returns only the matching lines plus configurable context (default 3 lines each side), capped at ~6 KB. Claude can locate the exact embed slot it needs with a single targeted search instead of loading the full source.

Token savings example: finding an \AddCard\ embed in a typical procedure - \search_embeditor_source(AddCard)\ returns ~20 lines (~400 tokens) vs \get_active_file\ returning the full generated source (~18,000 tokens). **~45x fewer tokens** for a targeted edit.

### \get_embed_content\
Reads the current code inside a specific embed slot by its line number (the N from \E:N\). Use this before rewriting an embed to avoid overwriting code you have not seen.

### \write_embed_content\
Writes Clarion code into an embed slot identified by line number. Writes through \PweePart.Data\ — the same property the embeditor itself uses — so the .app file is updated correctly. Returns a line delta; if non-zero, Claude knows to re-search before writing to subsequent embeds (line numbers shift when code is added/removed).

---

## Recommended workflow for Claude

search_embeditor_source(OPEN.Window)  - find the target slot (line N)
get_embed_content(N)                  - read existing code
write_embed_content(N, newCode)       - write new code
save_and_close_embeditor()            - persist to .app

This is documented as Critical Rule #9 in Terminal/clarion-assistant-prompt.md.

---

## Token efficiency summary

| Operation | Old approach | New approach | Saving |
|---|---|---|---|
| Find one embed slot | get_active_file ~18,000 tokens | search_embeditor_source ~400 tokens | ~45x |
| Read existing code | Manual parse of full source | get_embed_content exact slot only | ~100x |
| Write code | replace_range on generated text (fragile) | write_embed_content via PweePart.Data (safe) | No corruption risk |

---

## Files changed

- **Services/AppTreeService.cs** - Four new public methods + private SetIsDirty helper (walks AbstractViewContent inheritance chain via reflection to mark the .app file modified)
- **Services/McpToolRegistry.cs** - Four new tool registrations after the existing find_embed block
- **Terminal/clarion-assistant-prompt.md** - New PWEE Embeditor section with usage guidance; Critical Rule #9 for embeditor workflow
- **docs/embeditor-tools.md** - Reference documentation with token format details, recommended workflow, and implementation notes

---

## Notes

- SetIsDirty uses reflection to find the IsDirty property on AbstractViewContent (or a subclass) at runtime - consistent with the pattern used elsewhere in the codebase for SharpDevelop internals.
- Line numbers in E:N tokens are 1-based, consistent with all other MCP line-number conventions in this addin.
- The write_embed_content line delta mechanism ensures Claude self-corrects when editing multiple embeds in sequence - no stale line number bugs.

### DIFF
--- a/ClarionAssistant/Services/AppTreeService.cs
+++ b/ClarionAssistant/Services/AppTreeService.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Windows.Forms;
 using ICSharpCode.SharpDevelop.Gui;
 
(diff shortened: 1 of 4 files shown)


msarson:
## Bug found: `save_and_close_embeditor` triggers "Are you sure you want to save?" dialog

While testing this PR on my machine I noticed that `save_and_close_embeditor` causes Clarion to show a **"Are you sure you want to save?"** confirmation dialog, which then times out the MCP tool call (the tool blocks waiting for the dialog to be dismissed).

### Root cause

`SaveAndCloseEmbeditor()` calls `IGeneratorDialog.TryClose()`, which internally calls `BackToSource()` → `OnBackClick()`. When `IsDirty` is `true` (which it always is after `write_embed_content` sets it), `OnBackClick()` shows a `MessageBox.Show(...)` prompt. This is the interactive "Back" button handler — it was designed for human use, not programmatic use.

From the decompiled `CommonGenEditor`:

```csharp
private bool OnBackClick()
{
    if (IsDirty)
    {
        switch (MessageBox.Show("Save changes?", "Embed Editor?", YesNoCancel))
        {
            case DialogResult.Yes: ... Save(); ...
            // etc.
        }
    }
}
```

### Fix

`CommonGenEditor` has a public `SaveAndExit()` method that saves silently (no dialog) and then calls `CloseEditor()`:

```csharp
public void SaveAndExit()
{
    if (IsDirty)
        Save();       // silent — no MessageBox
    CloseEditor();
}
```

`SaveAndExit()` is not on `IGeneratorDialog` (that interface only exposes `TryClose`, `HaveChanges`, and `Discard`), so it needs to be called via reflection on the concrete editor object. The change in `SaveAndCloseEmbeditor()` would be:

```csharp
// Instead of:
var tryCloseMethod = dialogInterface.GetMethod("TryClose");
bool closed = (bool)tryCloseMethod.Invoke(editor, null);

// Use:
var saveAndExitMethod = editor.GetType().GetMethod("SaveAndExit",
    BindingFlags.Public | BindingFlags.Instance);
if (saveAndExitMethod != null)
{
    saveAndExitMethod.Invoke(editor, null);
    return "Embeditor saved and closed.";
}
// fallback to TryClose if SaveAndExit not found
```

Happy to push this fix to the branch if you'd like, or you can apply it yourself. Let me know.

peterparker57 (approved):
Thanks for this @msarson — really solid work. Token-efficient embed access with line-number-as-identifier plus the delta reporting on writes is exactly the right shape, and the noise-stripping in `get_embeditor_source` is a nice touch.

I'm going to merge and handle the polish items in a follow-up commit on master so we don't burn round-trip time on nits. For transparency, here's what I'll clean up:

**One real bug** — `McpJsonRpc.BuildSchema`'s dictionary values are parameter *descriptions*, not type hints. Existing callers pass human descriptions like `"Line number to go to (1-based)"`; this PR passes `"string"` / `"integer"`. Effect is cosmetic (the schema's parameter descriptions become the literal words "string" / "integer"; everything works at runtime) but I'll swap them out for real descriptions.

**Polish items:**
- `SetIsDirty` has a dead `StringBuilder log` parameter — the single call site passes `new StringBuilder()` that's immediately discarded. Will drop it or make it optional.
- Line-endings asymmetry: `PweePart.Data` gets LF-only, the document buffer gets CRLF. Will verify against what Clarion's native embeditor writes and align.
- `GetEmbeditorSource` prefixes embed lines with two spaces for display; `GetEmbedContent` does not. Risk: if Claude copies lines out of `search_embeditor_source` back into `write_embed_content`, the extra indent cascades. Will strip the prefix or document the split clearly.
- `WriteEmbedContentByLine` updates `PweePart.Data` before the document `Replace` — will reorder or add a rollback so a Replace failure can't leave model and view desynced.
- PR description says default `context_lines` is 3; code uses 5. Will align.
- `docs/embeditor-tools.md` says "SetIsDirty already exists in AppTreeService" — it's actually added by this PR. Minor doc fix.

Merging now. Thanks again!

msarson:
Hi John — following up on the webinar discussion about save_and_close_embeditor. You mentioned you might add a user option around the save behaviour — just wanted to check what you decided.

Our thinking: since save_and_close_embeditor is an MCP tool it's always called programmatically by Claude, never by a human. A confirmation prompt at that point doesn't protect the user — it just breaks the tool (the call blocks waiting for input that never comes, then times out).

The user's real safety gate is upstream — they've already reviewed and approved what write_embed_content proposed before asking Claude to save. By the time save_and_close_embeditor is called, the intent is clear.

So the fix feels straightforward: call SaveAndExit() via reflection instead of TryClose(), so the tool always saves silently. If a 'prompt' option is useful anywhere, it would make more sense on write_embed_content itself (e.g. a dry-run/preview flag) rather than on the save step.

Happy to push the SaveAndExit fix to the branch if that helps — just let us know what direction you want to go.

peterparker57:
Thanks for the ping @msarson — good catch that this slipped through the polish pass. Just shipped the fix in 8ed3868.

Went with your proposed approach: reflect for `SaveAndExit()` on the concrete editor type and call it directly, falling back to `IGeneratorDialog.TryClose` for older Clarion builds that don't expose the method. The fallback success message is tagged "via TryClose fallback" so we can tell when it fires in the wild.

One small addition on top of your sketch — since `SaveAndExit` is void and gives no signal if `Save()` fails silently, I re-read `IsDirty` after the call via a `GetIsDirty` helper (mirrors the existing `SetIsDirty` pattern, walks the inheritance chain to `AbstractViewContent`). If the flag is still true after the call, we return an error rather than claiming success, so the caller knows the editor may have been closed with unpersisted changes.

Tested end-to-end in the live IDE — save goes through silently, no modal, tool returns cleanly. Thanks again for the thorough writeup.

PR #20, opened 2026-04-22
Add GitHub Copilot CLI as alternative AI backend

Turpija12:
This adds GitHub Copilot CLI as a second AI backend alongside Claude Code, selectable per terminal session.

## What's new

- `Assistant.Backend` setting (Claude / Copilot) in Settings → Launch
- Copilot Commands — same Add/Edit/Delete UI as Claude commands; defaults: `copilot`, `gh copilot`, `copilot --allow-all-tools`
- Model selector for Copilot — dropdown passed via `--model` flag
- Permission Mode — Prompt or Allow Tools (`--allow-all-tools`)
- MCP auto-configured for Copilot with `tools: ["*"]` format
- Per-tab backend tracking so status bar works correctly when backend changes mid-session

## How to test

- Switch backend to Copilot in Settings → Launch, open a new tab — Copilot CLI should start
- MCP tools should work from within the Copilot session
- Claude backend should be unaffected

### DIFF
--- a/ClarionAssistant/ClarionAssistant.csproj
+++ b/ClarionAssistant/ClarionAssistant.csproj
@@ -192,6 +192,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Web.WebView2">
       <Version>1.0.3800.47</Version>
+      <GeneratePathProperty>true</GeneratePathProperty>
     </PackageReference>
   </ItemGroup>
   <Target Name="ValidateClarionReferences" BeforeTargets="ResolveReferences">
--- a/ClarionAssistant/ClarionAssistant.csproj
+++ b/ClarionAssistant/ClarionAssistant.csproj
@@ -52,6 +52,16 @@
       <HintPath>$(ClarionRoot)\bin\ICSharpCode.SharpDevelop.dll</HintPath>
       <Private>False</Private>
     </Reference>
+    <!-- Old-style csproj + PackageReference is inconsistent under dotnet/msbuild for this project.
+         Add explicit WebView2 references so compile-time assemblies are always resolved after restore. -->
+    <Reference Include="Microsoft.Web.WebView2.Core">
+      <HintPath>$(PkgMicrosoft_Web_WebView2)\lib\net462\Microsoft.Web.WebView2.Core.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="Microsoft.Web.WebView2.WinForms">
+      <HintPath>$(PkgMicrosoft_Web_WebView2)\lib\net462\Microsoft.Web.WebView2.WinForms.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
   </ItemGroup>
   <ItemGroup>
     <Compile Include="AgentPanelPad.cs" />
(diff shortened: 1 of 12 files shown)


peterparker57 (approved):
Thanks for the substantial contribution, @Turpija12. This adds a well-scoped second backend alongside Claude, with clean per-tab state (`AssistantLaunched`/`AssistantBackend`), a proper `McpConfigFormat` enum extension, and a working pwsh launch path that handles the x86-addin Program-Files resolution quirk. Build is clean under MSBuild Debug-C12, and a manual smoke test (Copilot CLI launches, MCP tools are reachable, Claude path unaffected, tab-exit label is backend-correct) passes end-to-end on my machine.

All four of the GitHub Copilot reviewer's suggestions are already addressed in commit 7ac3f46 (verified):
- `McpServer.GenerateMcpConfig` no longer builds the unused `toolNames` list for Copilot format
- `GeneratePathProperty="true"` is on the WebView2 `PackageReference`
- `ClaudeLaunched` → `AssistantLaunched` rename is complete (no stale references)
- `LaunchCopilotForTab` pwsh-missing path now disposes `Terminal` and resets flags

## Merging now — follow-up hotfix to land immediately after

I'm approving and squash-merging this to unblock the feature. I'll then push a small security/cleanup hotfix on master covering items that are either mine to own (pre-existing patterns that predate this PR) or small enough not to bounce the PR:

**Security hardening** (both pre-existing patterns, not introduced here, but the PR is a good moment to fix them):
- `ClaudeChatControl.cs:2570-2602` — `Copilot.ExtraFlags` is concatenated into the pwsh `-Command` string without quoting, while every other user-controllable value in that method uses `.Replace("'", "''")` + single-quote wrap. Same is true for `Copilot.Commands` entries that aren't the bare string `"copilot"`. A settings-file writer (e.g., a malicious config snippet a user is told to paste) could inject arbitrary pwsh. Same shape exists on the Claude side (`Claude.ExtraFlags` / `claudeBase`). Fix: switch to `ProcessStartInfo.ArgumentList` or enforce per-token quoting for both backends.
- `SettingsService.Save/Load` — `key=value\n` format with no newline rejection on values. A value containing `\n` smuggles a second key on reload (e.g. lets an attacker escalate `Copilot.PermissionMode` to `allow`). Fix: reject/escape `\n`/`\r`/`=` in values.

**Quality cleanup:**
- Per-tab `%AppData%\ClarionAssistant\copilot\tab-<id>` dirs accumulate — never cleaned up. Will collapse to a single shared dir since contents are launch-identical per tab.
- One-line comments on the `'@<path>'` MCP config syntax and the `--add-dir` duplication with the `cd` CWD (non-obvious to future readers).
- `IsNullOrEmpty` → `IsNullOrWhiteSpace` on `Copilot.ExtraFlags`.

## Follow-up issues I'll file separately

- MCP HTTP server has `Access-Control-Allow-Origin: *` with no auth token — this PR widens the exposed tool surface via `tools: ["*"]`, but the CORS gap is pre-existing. Browser drive-by → local RCE is the worst-case. Needs a per-session token in the MCP config + strict origin check.
- Rename `ClaudeChatControl` → `AssistantChatControl` (large blast-radius, dedicated PR)
- Extract shared launch scaffolding before a third backend — `LaunchClaudeForTab` and `LaunchCopilotForTab` now share ~30 lines of scaffolding verbatim
- Merge `Terminal/AGENTS.md` with `Terminal/clarion-assistant-prompt.md` (or cross-reference) so the two sources don't drift

None of these block merging. Thanks again for the work — merging shortly.

PR #29, opened 2026-05-13
Terminal UX: context menu, smart copy, large paste as @file

bdinko:
## Summary

Three independent quality-of-life improvements to the embedded terminal, all inside `Terminal/` — no changes to the Claude/Copilot/Codex launch paths.

### Context menu

`terminal.html` already posted a `contextmenu` message on right-click, but no C# handler existed, so right-click in a Claude tab was silently dropped. Added a WinForms `ContextMenuStrip` with:

- **Copy** — flowed text (paragraphs joined; see "Smart copy" below).
- **Copy raw (no paragraph join)** — only shown when raw differs from flowed (i.e. code blocks, tables, structured output where joining would damage layout).
- **Paste** — same path as Ctrl+V.
- **Paste from file…** — `OpenFileDialog`, injects `@<path>` into the terminal.
- **Paste large clipboard as file** — only shown when the clipboard exceeds the large-paste threshold.

### Large-paste-as-`@file`

Claude Code's prompt UI caps at ~5-6 visible rows; pasting a long blob makes the prompt unreadable and unscrollable. `OnPasteRequested` now routes any clipboard text over the threshold (>1000 chars or >4 newlines) through a temp-file path instead of streaming raw bytes into ConPTY:

- Writes the text to `%LOCALAPPDATA%\ClarionAssistant\paste-cache\paste-<utc-timestamp>.txt` as UTF-8 without BOM.
- Prunes `paste-*.txt` files older than 24h on each write.
- Calls `GetShortPathName` (new P/Invoke in `NativeMethods`) to dodge spaces in `C:\Users\<First Last>\…` that would break Claude's whitespace-terminated `@reference` parser.
- Injects `@<short-path> ` into ConPTY — Claude reads the file on submit and the prompt stays visually compact. Trailing space terminates the reference and leaves the cursor positioned for the user's question.
- If cache write fails, falls back to the original raw-paste behavior so the user's clipboard isn't silently lost.

### Smart copy (Ctrl+C and right-click Copy)

`xterm.js`'s `getSelection()` returns each row padded with cell whitespace to the terminal width and emits a hard `\n` at every visual wrap. For Claude's Ink-rendered output, every wrapped row is also a discrete buffer line (`isWrapped` is always `false`) with explicit leading alignment spaces — so soft-wrap detection alone doesn't help.

New `smartGetSelection` reads the buffer rows directly via `buffer.active.getLine(y).translateToString` and offers two modes:

- **raw**: one logical line per visual row, trailing cell padding stripped. The "Copy raw" menu item.
- **flowed**: rows that start with whitespace are treated as continuations of the previous paragraph and joined into the previous logical line (the leading indent is what Ink-based TUIs write on wraps). Blank rows become paragraph breaks. This is the default for Ctrl+C and right-click Copy.

A three-paragraph prose response now arrives in the clipboard as three lines separated by blank rows, matching what the user sees on screen. Structured output (code, tables) keeps its line breaks via Copy raw.

## Test plan

- [x] Right-click in a Claude tab — context menu appears.
- [x] Select prose, Ctrl+C → Notepad: paragraphs are flowed, no trailing column padding, no leading alignment indent.
- [x] Same selection via right-click → Copy: same result.
- [x] Right-click → Copy raw (no paragraph join): trailing padding still stripped, but every visual row preserved as its own line.
- [x] Paste from file… → choose a multi-KB text file → prompt receives `@<short-path>` instead of file contents; Claude reads the file on submit.
- [x] Ctrl+V with multi-KB clipboard payload → auto-routed through the same `@file` path.
- [x] Ctrl+V with short clipboard → behaves exactly as before (raw bytes to ConPTY).
- [x] User profile with spaces in the path → injected `@<short-path>` resolves correctly.
- [x] Restarting Clarion clears stale paste-cache entries older than 24h.

### DIFF
--- a/ClarionAssistant/Terminal/NativeMethods.cs
+++ b/ClarionAssistant/Terminal/NativeMethods.cs
@@ -99,6 +99,11 @@ namespace ClarionAssistant.Terminal
 
         internal const uint HANDLE_FLAG_INHERIT = 0x00000001;
 
+        // 8.3 short path resolution — used to sidestep spaces in user-profile paths
+        // when injecting "@<file>" references into Claude Code's prompt parser.
+        [DllImport("kernel32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
+        internal static extern uint GetShortPathName(string lpszLongPath, System.Text.StringBuilder lpszShortPath, uint cchBuffer);
+
         #region Job Object API
 
         [DllImport("kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
--- a/ClarionAssistant/Terminal/WebViewTerminalRenderer.cs
+++ b/ClarionAssistant/Terminal/WebViewTerminalRenderer.cs
@@ -173,6 +173,9 @@ namespace ClarionAssistant.Terminal
                     case "paste":
                         OnPasteRequested();
                         break;
+                    case "contextmenu":
+                        OnContextMenuRequested(message.X, message.Y, message.SelectedText, message.SelectedTextRaw);
+                        break;
                     case "fontSizeChanged":
                         float newSize;
                         if (float.TryParse(message.Data, out newSize))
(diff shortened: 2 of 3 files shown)


bdinko:
Rebased this branch onto current master (v4.6.2) — clean, no conflicts (the changes are isolated to `Terminal/NativeMethods.cs`, `Terminal/WebViewTerminalRenderer.cs`, and `Terminal/terminal.html`, none of which master touched in 4.6.x).

Built and deployed against **Clarion 11.1 on v4.6.2** and confirmed the prompt-line right-click context menu (and smart copy / large-paste-as-@file) works as expected. Ready for review.

peterparker57 (approved):
Approved — solid, well-documented UX work. Thanks @bdinko.

Verified locally: built the PR head in an isolated worktree against current `master` — **0 errors**, `ClarionAssistant.dll` produced (C12 config). All `using`s are present for the new WinForms/IO/Text APIs, threading is correct (context menu / clipboard / OpenFileDialog all driven from the UI-thread `OnWebMessageReceived`), and the large-paste short-path P/Invoke is correctly gated to only fire when the path contains a space. Defensive `catch {}` + raw-paste fallback matches the file's "never crash the IDE host" style.

Two non-blocking follow-ups for later (not merge blockers):

1. **Hand-rolled JSON parser + `\u` escapes.** `ParseJsonString` handles `\n \r \t \" \\` but its `default` case drops the backslash for `\uXXXX`. This PR is the first to route freeform `selectedText` through it. Impact is small (modern `JSON.stringify` only `\u`-escapes control chars — already stripped from selections — plus U+2028/2029 and lone surrogates), but those would corrupt on copy. The `input` path already sidesteps this by base64-encoding; `selectedText` could do the same.
2. **`ContextMenuStrip` not disposed.** A new one is created per right-click and never disposed — `menu.Closed += (s,e) => menu.Dispose();` would fix the minor leak.

Edge note: if 8.3 short-name generation is disabled on the volume, `GetShortPathName` returns the long path unchanged, so a spaced cache path would degrade to a broken `@ref`. Rare config.

Merging.

PR #35, opened 2026-05-23
deploy.ps1: auto-discover MSBuild instead of hardcoding VS2022 path

bdinko:
## Problem

`ClarionAssistant/deploy.ps1` hardcodes the MSBuild path to VS 2022 Community:

```powershell
$MSBuild = "C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Current\Bin\MSBuild.exe"
```

This breaks the deploy on any machine that doesn't have that exact edition/version — e.g. VS 2019, Professional/Enterprise editions, or the newer VS "18" line where MSBuild lives under `...\Visual Studio\18\Community\...`. The script fails immediately at the build step with a path-not-found error.

## Fix

Replace the literal path with a `Resolve-MSBuild` function that:

1. Queries `vswhere -latest -requires Microsoft.Component.MSBuild` — covers any installed VS edition/version.
2. Falls back to scanning Program Files for `Current\Bin\MSBuild.exe` if `vswhere` is unavailable.
3. Throws a clear, actionable error if MSBuild genuinely can't be found.

No behavior change on machines that already worked; it just stops assuming a single hardcoded location.

## Testing

Verified `Resolve-MSBuild` resolves correctly on a machine where the old hardcoded path does **not** exist — it located MSBuild under `...\Visual Studio\18\Community\...` via vswhere, and `deploy.ps1 -Version 12` built `ClarionAssistant.dll` with 0 warnings / 0 errors.

### DIFF
--- a/ClarionAssistant/deploy.ps1
+++ b/ClarionAssistant/deploy.ps1
@@ -11,9 +11,32 @@ param(
 
 $ErrorActionPreference = "Stop"
 
+# Locate MSBuild.exe without hardcoding a Visual Studio version/edition.
+# Order: vswhere (covers VS 2019/2022/18+, any edition) -> common install paths.
+function Resolve-MSBuild {
+    $vswhere = Join-Path ${env:ProgramFiles(x86)} "Microsoft Visual Studio\Installer\vswhere.exe"
+    if (Test-Path $vswhere) {
+        $found = & $vswhere -latest -requires Microsoft.Component.MSBuild `
+                            -find "MSBuild\**\Bin\MSBuild.exe" | Select-Object -First 1
+        if ($found -and (Test-Path $found)) { return $found }
+    }
+
+    # Fallback: scan known roots if vswhere is unavailable.
+    foreach ($root in @($env:ProgramFiles, ${env:ProgramFiles(x86)})) {
+        if (-not $root) { continue }
+        $candidate = Get-ChildItem -Path (Join-Path $root "Microsoft Visual Studio") `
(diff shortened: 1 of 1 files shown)


bdinko:
Tested on a machine where the old hardcoded VS2022 Community path does **not** exist: `Resolve-MSBuild` correctly located MSBuild via vswhere (under `...\Visual Studio\18\Community\...`), and `deploy.ps1` then built and deployed the addin with 0 warnings / 0 errors against Clarion 10, 11.1, and 12. No behavior change on setups that already worked. Ready for review.

peterparker57 (approved):
@
Approved — clean, well-scoped fix. Thanks @bdinko.

Verified against the full `deploy.ps1`: the new `Resolve-MSBuild` slots in correctly at the `$MSBuild` assignment and is consumed unchanged by the Restore/Build/Indexer steps. The vswhere call (`-latest -requires Microsoft.Component.MSBuild -find "MSBuild\**\Bin\MSBuild.exe"`) is the canonical discovery pattern and resolves the *same* 32-bit `Current\Bin\MSBuild.exe` the old hardcoded path pointed at, so no architecture/behavior drift on machines that already worked. Fallback regex correctly excludes the `amd64`/`Roslyn` copies, and the `throw` message is valid.

Two optional, non-blocking follow-ups for later:
1. The Program Files fallback `Get-ChildItem -Recurse` walks the whole VS tree — only runs when vswhere is absent (rare), but could be tightened with `-Depth`/a narrower path.
2. Consider adding `-prerelease` to the vswhere call so a machine with only a VS Preview install resolves on the fast path instead of falling through to the scan.

Merging.
@

PR #37, opened 2026-06-05
Fix LSP redirection handshake and make path resolution config-aware

msarson:
## Problem

The addin acts as the host/client for the Clarion language server (from the Clarion VS Code extension), spawning it over stdio and driving the LSP handshake. The `clarion/updatePaths` notification was being sent in the wrong shape, so the server could not locate the project's `.red` redirection file. It fell back to the install-default redirection, flooded the log with `No valid redirection file found` (×40+), and loaded **zero** projects — leaving `lsp_` tools with no data.

### Root cause

The server resolves the effective `.red` itself:

```js
path.join(projectPath, redirectionFile)            // project-local override
path.join(redirectionPaths[0], redirectionFile)    // install/global fallback
```

So the payload contract is:

- **`redirectionFile` must be a bare filename** (e.g. `Clarion100.red`), **not** an absolute path. An absolute path makes `path.join` produce a non-existent target (`path.join` doesn't reset on an absolute second segment), so neither the project nor the global red is ever found.
- **`redirectionPaths[0]` must be the reddir directory** (where the install/global red lives); the per-project `.red` is discovered via `projectPaths[0]` (the solution dir).

The addin was sending an absolute install path as `redirectionFile` and a list of `.clw` source dirs as `redirectionPaths`.

## Fix

**1. Correct handshake shape.** New `McpToolRegistry.BuildLspUpdatePaths()` emits the shape the server expects (`RedFileName` + reddir directory), and is used from **both** start paths:
- `EnsureLspRunning()` (auto-start), and
- the `lsp_start` tool — which previously sent **no** `updatePaths` at all, so a manual start left the server with no redirection info.

**2. Config-aware path resolution via delegation.** The local `RedFileService` only searches the `Common` section, so it missed config-specific paths (e.g. `.\genfiles\src` declared under `[Debug]`). The server is already config-correct (it merges the active configuration section with `Common`). Rather than duplicate that logic, path resolution now delegates to the server:
- Added `LspClient.FindFile` (`clarion/findFile`) and `LspClient.GetServerSearchPaths` (`clarion/getSearchPaths`).
- `resolve_red_path` and `get_red_search_paths` now use the server's config-aware resolver, falling back to the local `Common`-only `RedFileService` only when the LSP is down or the solution hasn't finished loading.
- Responses are tagged `resolver: lsp | redfile-fallback` for transparency.

## Testing

- Builds clean (Debug, 0 warnings / 0 errors).
- Verified in a Clarion 10 VM: the project `.red` now parses, the `No valid redirection file found` flood is gone, and all projects load. `resolve_red_path` returns `resolver: lsp` and resolves config-specific paths that the `Common`-only lookup previously missed.

## Notes / follow-ups

- First `resolve_red_path` after a cold start may fall back to `RedFileService` until the server finishes building the solution; subsequent calls hit the LSP.
- `clarion/getClwDirectories` delegation (for the generated-`.clw` locator) is not wired yet — left as a follow-up.

### DIFF
--- a/ClarionAssistant/Services/LspClient.cs
+++ b/ClarionAssistant/Services/LspClient.cs
@@ -343,6 +343,84 @@ namespace ClarionAssistant.Services
             return SendRequest("textDocument/rename", parms, 8000);
         }
 
+        /// <summary>
+        /// clarion/findFile — ask the server to resolve a filename to its absolute path using
+        /// the server's full, config-aware redirection logic (active configuration section +
+        /// Common, with libsrc fallback). This is the authoritative resolver: prefer it over any
+        /// local Common-only redirection lookup, which misses config-specific paths like
+        /// .\genfiles\src.
+        ///
+        /// Returns the inner { path, source } result dictionary, or null when the server is down,
+        /// no solution is loaded server-side, or the file isn't found.
+        /// </summary>
+        public Dictionary<string, object> FindFile(string filename, string sourceUri = null)
+        {
+            if (!IsRunning || string.IsNullOrEmpty(filename)) return null;
+            TrackRequest("findFile", filename);
+
+            var parms = new Dictionary<string, object> { { "filename", filename } };
+            if (!string.IsNullOrEmpty(sourceUri)) parms["sourceUri"] = sourceUri;
+
+            var response = SendRequest("clarion/findFile", parms);
+            var result = ExtractResultObject(response) as Dictionary<string, object>;
+            if (result == null) return null;
+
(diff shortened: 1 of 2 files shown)


msarson:
For broader context: this PR fixes one instance of a recurring pattern — the addin running its own (Common-only) resolution in parallel to the language server. I've catalogued the other spots where delegating to the LSP would help in #38 (discussion/opportunities, not a demand for PRs). This PR stands on its own; #38 is just the bigger picture.

Issue #44, opened 2026-06-22
apply_embed_edits proposal + non-ASCII (Windows-1252) corruption in embed writes


Aarhusdk:
Hi John,

Two things bundled together — a small tool we'd like to propose, and a character-encoding bug we hit while using it that probably affects the existing embed-write path too.

## 1. Proposed tool: `apply_embed_edits`

We added a tool for editing **very large procedures** where driving the classic PWEE editor interactively (repeated `open_procedure_embed` + `write_embed_content`) destabilises and crashes the IDE. Real case: a ~4,500-line `Touch_Salg` procedure in a POS app.

It does **one transient** `open → write → save → close` round-trip with no interactive session left open, reusing the proven Modern Embeditor save path:

`ModernEmbeditorLauncher.OpenAndMirror → WriteEmbedContentByLine (bottom-to-top) → SaveAndCloseEmbeditor → WaitForEmbedClosed`

- Input: `procedure_name` + `edits` = JSON array of `{ "line_number": N, "code": "...full slot replacement..." }`, where `N` is the 1-based `«E:N»` slot start.
- **All-or-nothing:** every `line_number` is validated against the freshly-opened embed structure first; if any is not a current slot start, nothing is written.
- Writes apply bottom-to-top so earlier slots' line numbers stay valid.

In practice this let us apply 5 multi-line edits across a huge procedure with zero IDE crashes, where the interactive path reliably crashed. Happy to send a PR if you'd like it upstream.

## 2. Bug: non-ASCII text is double-encoded on write (UTF-8 → Windows-1252)

Clarion source is **Windows-1252 / ANSI**. When the `code` payload contains non-ASCII characters, they come out **mojibaked** in the saved source:

| Sent | Saved |
|------|-------|
| `på` | `pÃ¥` |
| `Åben` | `Ã…ben` |
| `håndtering` | `hÃ¥ndtering` |

This is the classic UTF-8-bytes-interpreted-as-Windows-1252 double-encoding. Because `ApplyLineEdits` routes through the same `WriteEmbedContentByLine` as the interactive `write_embed_content`, **the existing embed-write path is likely affected too** for any non-ASCII content.

**Likely root cause:** an encoding mismatch between (a) how the MCP JSON-RPC request body is decoded and (b) the ANSI/Windows-1252 encoding Clarion expects on disk — i.e. the incoming UTF-8 request bytes aren't decoded as UTF-8, or the write isn't transcoded back to the source file's ANSI code page.

[...]

PR #45, opened 2026-06-22
feat(mcp): add apply_embed_edits for crash-free edits to large procedures

Aarhusdk:
## What

Adds an MCP tool **`apply_embed_edits`** that applies one or more embed-slot edits to a procedure in a **single transient** `open → write → save → close` round-trip, with no interactive embeditor session left open.

## Why

Driving the classic PWEE editor interactively — repeated `open_procedure_embed` + `write_embed_content` — destabilises and crashes the IDE on **very large procedures**. Real case: a ~4,500-line `Touch_Salg` procedure in a POS app, where the interactive path crashed reliably. This tool let us apply 5 multi-line edits across that procedure with **zero crashes**.

It reuses the proven Modern Embeditor save path rather than inventing a new one:

```
ModernEmbeditorLauncher.OpenAndMirror
  → WriteEmbedContentByLine (bottom-to-top)
  → SaveAndCloseEmbeditor
  → WaitForEmbedClosed
```

## How

- **`ModernEmbeditorSaver.ApplyLineEdits(procName, edits, out ok)`**
  - Validates every edit's `line_number` against the freshly-opened embed structure **before writing anything** (all-or-nothing).
  - Writes changed slots **bottom-to-top** so earlier slots' line numbers stay valid.
  - On any failure it cancels the embeditor and persists nothing.
- **`McpToolRegistry`** registers the `apply_embed_edits` tool.
  - Input: `procedure_name` + `edits` = JSON array of `{ "line_number": N, "code": "...full slot replacement..." }`, where `N` is the 1-based `«E:N»` slot start (from `get_embeditor_source` / `search_embeditor_source`).

## Notes

- Scope is intentionally just the two service files — no version bump or deploy-script changes are included.
- Related: #44 documents a non-ASCII (UTF-8 → Windows-1252) double-encoding issue observed via this same write path (`WriteEmbedContentByLine`), which likely affects the existing interactive `write_embed_content` too. That's a separate fix; this PR only adds the tool.

Happy to adjust naming/placement to fit your conventions.

### DIFF
--- a/ClarionAssistant/Services/McpToolRegistry.cs
+++ b/ClarionAssistant/Services/McpToolRegistry.cs
@@ -980,6 +980,58 @@ Use this tool to discover IDE APIs and understand what's available for automatio
                 }
             });
 
+            Register(new McpTool
+            {
+                Name = "apply_embed_edits",
+                Description = "Apply one or more embed-slot edits to a procedure in a SINGLE transient " +
+                    "open->write->save->close round-trip — NO interactive embeditor session stays open. " +
+                    "Prefer this over open_procedure_embed + write_embed_content for LARGE procedures where the " +
+                    "live PWEE editor is unstable under repeated interactive driving (it reuses the proven Modern " +
+                    "Embeditor save path). 'edits' is a JSON array of {\"line_number\":N,\"code\":\"...\"}: line_number " +
+                    "is the 1-based «E:N» slot start (from get_embeditor_source/search_embeditor_source), code is the " +
+                    "COMPLETE replacement for that slot (end with a trailing newline). Edits are applied bottom-to-top. " +
+                    "If ANY line_number is not a current embed-slot start, NOTHING is written. The procedure is opened, " +
+                    "written, saved and closed automatically — do NOT wrap this in open_procedure_embed / " +
+                    "save_and_close_embeditor.",
+                InputSchema = McpJsonRpc.BuildSchema(new Dictionary<string, string>
+                {
+                    { "procedure_name", "Name of the procedure to edit." },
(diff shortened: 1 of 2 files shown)


peterparker57:
Reviewed — this looks good and I'd recommend merging it. 🚀

**Logic is sound.** `ApplyLineEdits` is a faithful sibling of the existing, proven `ModernEmbeditorSaver.Save` path. It correctly drops the Monaco snapshot-equality checks (no snapshot to compare against here — edits arrive directly) while keeping every safety property that matters:

- **All-or-nothing validation** — every `line_number` is checked against the freshly-opened slot starts *before* any write; one bad line cancels the embeditor and persists nothing.
- **Bottom-to-top writes** (`OrderByDescending(x => x.Key)`) so earlier slots' line numbers stay valid — matches `Save` exactly.
- **Cancel-on-failure** for partial write errors, exceptions, and the unconfirmed-close case.
- Reuses the vetted `OpenAndMirror → WriteEmbedContentByLine → SaveAndCloseEmbeditor → WaitForEmbedClosed` chain rather than inventing a new one. `RequiresUiThread = true` is correct.

**Minor, non-blocking nits:**
- No de-dup of `line_number`s — two edits targeting the same slot would both write (last-applied bottom-to-top wins). Edge case only.
- We'll add a version bump on merge so deployers can tell it shipped (you intentionally left that out — thanks for keeping the PR tight).

**On the #44 encoding bug:** confirmed and root-caused — see my comment on #44. You were right that it's a separate concern from this PR, and right that it lives in the shared write path. We're fixing it on our side in `McpServer.cs` (the request-body decode, not the write path itself), so once that lands you can drop the ASCII-only `æ ø å → ae oe aa` workaround.

Thanks for the clean, well-documented contribution! 🙏

PR #53, opened 2026-07-02
Add Code Snippets (Ctrl+J) to the CA Embeditor

### LINKED ISSUE #49: Request: Code Snippets / Templates support in CA Embeditor (Ctrl+J style)
## Summary

Request to add a code snippet/template system to the CA Embeditor, similar to the 
classic Clarion IDE's "Code Templates" feature (Tools > Options > Coding), 
triggered via Ctrl+J.

## Background / Motivation

The classic Clarion Text Editor (since Clarion 7) has a built-in snippet system 
often referred to as "Code Templates" (distinct from Application Generator 
templates). Workflow:

1. Developer types a trigger word (e.g. `Ventas`)
2. Presses Ctrl+J
3. A popup list appears showing snippets registered for that trigger/context
4. Selecting one inserts a pre-defined code block (e.g. a full LOOP construct 
   for the `Ventas` file, including OPEN/CLOSE, key selection, etc.)

This is configured per file extension in Tools > Options > Coding, and is one of 
the most-used productivity features for developers with large, established 
codebases who repeat structural patterns constantly (loops over specific files, 
standard error handling blocks, standard header comments, etc.).

Since CA Embeditor is a separate editor (Monaco/WebView2-based) from the classic 
IDE Text Editor, this snippet feature is not currently available there, which 
is a regression in workflow for developers migrating to CA Embeditor as their 
primary editor.

## Current State

- CA Embeditor is built on Monaco, which is the same editor engine VS Code uses.
- Monaco has native, first-class support for a "Snippet" completion item kind 
  via `monaco.languages.registerCompletionItemProvider`, including tab-stop 
  placeholders (`$1`, `$2`, `${1:default}`) and variable substitution — this is 
  the exact mechanism VS Code's own user snippets are built on.
- As far as I can tell from the README, changelog, and MCP tool list, this 
  capability exists in the underlying editor but is not yet exposed as a 
  feature by ClarionAssistant.

## Proposed Feature

1. **Snippet storage**: A simple JSON or XML file (e.g. 
   `%APPDATA%\ClarionAssistant\snippets.json`) storing snippet definitions:
   - Trigger keyword
   - Optional description
   - Target file extension(s)/context (.clw, embed points, etc.)
   - Snippet body, with Monaco-style tab-stop/placeholder syntax so cursor 
     lands in the right spot after insertion (e.g. jumping into a filter 
     condition)

2. **Snippet editor UI**: A settings page (similar to the existing "Most users 
   can skip this page" MCP Server settings pattern) to Add/Edit/Delete 
   snippets — mirroring the old Tools > Options > Coding dialog closely enough 
   that existing Clarion devs feel at home.

3. **Trigger**: Ctrl+J (matching the classic IDE shortcut, for muscle-memory 
   continuity) opens a filtered popup of snippets matching the word currently 
   under/before the cursor, inserted at cursor position, with tab-stops.

4. **Optional stretch goal**: Import existing Code Template definitions from 
   the classic IDE's config (if they're stored in an accessible format) so 
   existing snippet libraries carry over without manual re-entry.

## Why this matters

This would remove one of the last workflow gaps preventing devs from using 
CA Embeditor as a full daily-driver replacement for the classic Text Editor, 
which is presumably the long-term goal of the CA Embeditor effort based on the 
v5.0 roadmap notes.

## Willingness to contribute

Happy to help test, provide sample snippet sets from real-world usage, or 
contribute towards implementation if there's appetite for it and guidance on 
where in the codebase the Monaco integration lives.

asantarelli:
## Summary
- Replicates the classic Clarion text editor's Ctrl+J template picker on the Monaco-based CA Embeditor, using Monaco snippet syntax for tab-stops.
- Ctrl+J opens a dedicated filterable picker (not folded into the LSP completion provider), scoped to the active file's extension.
- Snippet bodies support `$1` / `${1:default}` / `$0` tab-stops plus a custom `${SELECTED}` placeholder that's substituted with whatever text was selected when Ctrl+J was pressed (falls back to a fillable tab-stop when nothing was selected). Continuation lines auto-indent to match the line Ctrl+J was triggered on.
- New "Snippets" page in Settings (table + add/edit/delete modal, mirrors the existing GitHub Accounts page) backed by a new `SnippetStore` — global JSON storage at `%APPDATA%\ClarionAssistant\snippets.json`. Changes broadcast live to every open embeditor tab.
- Snippet insertion goes through `executeEdits` with a small home-grown tab-stop engine rather than Monaco's native `editor.action.insertSnippet` — that action's edits aren't recognized by `installReadOnlyGuard` as a legal single edit inside embed slots (same reason the Data-pad drop path already avoids it), so it silently gets undone there.

Closes #49.

## Test plan
- [x] Manually verified in a running Clarion 11 IDE: create/edit/delete snippets in Settings > Snippets.
- [x] Ctrl+J in the CA Embeditor (procedure/embed mode) opens the picker, filters live, inserts with tab-stops and `${SELECTED}` substitution working with and without an active selection.
- [x] Tab/Shift+Tab cycles between remaining tab-stops after insertion.
- [x] Continuation-line indentation matches the trigger line.
- [x] Editing snippets in Settings while an embeditor tab is open updates that tab's picker live (broadcast).
- [x] Extension filtering (`.clw`-only snippet doesn't show while editing a `.tpl`, etc.).
- [ ] Reviewer: exercise Ctrl+F/Ctrl+H/F12/Ctrl+D still work normally and don't fire while the snippet picker is open.

### DIFF
--- a/ClarionAssistant/ClarionAssistant.csproj
+++ b/ClarionAssistant/ClarionAssistant.csproj
@@ -107,6 +107,7 @@
       <DependentUpon>ClassHelperControl.cs</DependentUpon>
     </Compile>
     <Compile Include="Services\ClarionClassParser.cs" />
+    <Compile Include="Services\SnippetStore.cs" />
     <Compile Include="Services\ExplorerRecentsStore.cs" />
     <Compile Include="Services\ExplorerFileClassifier.cs" />
     <Compile Include="Services\MonacoFileOpener.cs" />
--- a/ClarionAssistant/Dialogs/ClaudeChatSettingsDialog.cs
+++ b/ClarionAssistant/Dialogs/ClaudeChatSettingsDialog.cs
@@ -187,6 +187,15 @@ namespace ClarionAssistant.Dialogs
                     case "testGitHubToken":
                         HandleTestGitHubToken(json);
                         break;
+                    case "addSnippet":
+                        HandleAddSnippet(json);
+                        break;
+                    case "editSnippet":
+                        HandleEditSnippet(json);
+                        break;
+                    case "deleteSnippet":
+                        HandleDeleteSnippet(json);
+                        break;
                     case "browseClassFolder":
                         BrowseFolder("classFolder", "Select Class Output Folder", _settings.Get("Class.OutputFolder"));
                         break;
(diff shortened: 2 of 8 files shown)


ClarionLive:
Thanks again @asantarelli — this landed cleanly and then grew quite a bit on top of your foundation. Documenting the follow-up here for traceability (all on `master`, commits `3644b9e … ff9a56a`, shipped through v5.1.660):

**Relocation**
- Moved the snippet manager out of the Clarion Assistant Settings dialog and into the **CA Embeditor's gear panel** (Editor | Code Snippets tabs) — snippets are an editor concern, not an app-settings one. `SnippetStore` + `%APPDATA%\ClarionAssistant\snippets.json` unchanged; parsing was lifted into a shared `SnippetStore.ApplyCommand` so both editor hosts write through one path.

**Insertion — classic Clarion parity**
- Snippets now appear in the **live completion list** (snippet icon + a body preview): type a trigger prefix → **Enter/Tab** expands.
- **`trigger + <space>`** inline expansion via a dedicated space-key handler (Monaco doesn't run a completion item's command on a space-commit, only Enter/Tab — so the space key owns it).
- **Ctrl+J** picker retained for browsing; it's now theme-aware (light/dark), has X / Select / Cancel chrome, and a two-column list + body preview.
- `${SELECTED}` snippets are **Ctrl+J-only** — you can't hold a selection while typing a trigger, so they're kept out of the completion/space path.

**Tab-stop engine**
- Added **mirroring** for repeated `${N}` (type the table name once → `OPEN/SET/NEXT/CLOSE` all fill).
- Fixed the two items from the original review of this PR: the hardcoded `seleccionar` default → `selected`, and an **editable-range gate** so Ctrl+J in generated/read-only code no longer lets the read-only guard revert the insert while leaving phantom tab-state.

**Polish**
- **Case-insensitive** extension scoping *and* triggers (no `.CLW;.Clw;.clw` needed).
- An examples **"?"** help dialog in the snippet form (worked examples, before/after).

A multi-agent review pass caught a few mirror-engine bugs before release — non-primary occurrence edits being discarded, `$0` not ending the snippet, and Ctrl+S-before-Tab persisting un-converged mirrors — all fixed and re-verified.

Genuinely appreciated how well you studied the codebase — deliberately avoiding Monaco's native `insertSnippet` for the read-only-guard reason saved us a real headache, and we kept that approach throughout. Closes #49. 🙏

Issue #2, opened 2026-04-03
replace_text destroys PWEE embeditor content — edits don't persist


msarson:
When using `replace_text` inside the PWEE embeditor (app generator embed editor), edits appear to work but don't save back to the `.app` file, and subsequent operations produce errors like `index should be between 0 and 42`.

**Root cause:** `EditorService.ReplaceText` sets `Document.TextContent` on the full document. In the PWEE embeditor the document contains *all* embed points interleaved with read-only generated sections tracked by `PweeLineManager`. Replacing `TextContent` wipes everything and leaves `PweeLineManager`'s `StartLineNr`/`EndLineNr` indices pointing at stale positions — the dirty flags never get set correctly and `Save()` silently skips the embed.

**`ReplaceRange` is fine** — it uses `Document.Replace(offset, length, text)` surgically, which fires `DocumentChanged` → `MoveIndices`, keeping all line indices correct.

**Fix approach:**
- Get `currentEmbed` (private field on `CommonGenEditor`, kept current by `Caret_PositionChanged` after any navigation call)
- Get the embed's `StartLineNr`/`EndLineNr` from that `CustomPweeLine`
- Use `Document.Replace(startOffset, length, newText)` on only that line range
- `PweeLineManager.DocumentChanged` then auto-shifts all other embeds and marks the modified one `Dirty = true`

Also worth noting: `Caret` is not directly on `TextEditorControl` — the correct path is `TextEditorControl.ActiveTextAreaControl.Caret.Line`.

peterparker57:
This has been fixed. `ReplaceText` now uses `Document.Replace(offset, length, text)` surgically (replacing end-to-start to preserve offsets) instead of setting `TextContent` wholesale. This fires `DocumentChanged` so `PweeLineManager.MoveIndices` keeps embed tracking correct. The caret path also correctly goes through `ActiveTextAreaControl.TextArea`. The `TextContent` setter only remains as a fallback for editors that lack the `Replace` method.

PR #57, opened 2026-07-05
fix(mcp): prevent HTTP listener deadlock when UI thread is busy

Aarhusdk:
## Problem

The streamable-HTTP MCP listener can wedge permanently: Clarion.exe still owns the http.sys queue with the URL registered, but no requests are dequeued — `POST /mcp initialize` hangs >30 s (HTTP 000) and a CLOSE_WAIT socket is left behind. Only an IDE restart recovers. Observed in the field on 5.2.691.

Root cause: UI-thread tool dispatch uses a **synchronous** `Control.Invoke`. When the IDE UI thread is busy/blocked, the worker thread waits forever, holding the request open and leaking the connection as CLOSE_WAIT. With a few of these piled up, the listener stops serving everything — including `initialize`.

## Fix (3 parts)

1. **Bounded UI-thread dispatch** — `Control.Invoke` replaced with `BeginInvoke` + `ManualResetEventSlim.Wait(30 s)`. On timeout the request returns a proper tool error (`UI thread did not respond within 30s`) instead of hanging the worker; the abandoned delegate completes harmlessly later.
2. **Always close the response** — `HandleMessagePost` had early-return paths and no exception guard around `response.Close()`, so failed requests left connections half-open. The handler body is now wrapped in `try/finally` that always closes the response.
3. **ThreadPool warm-up** — min worker threads raised to 16 at `Start()` so a few blocked UI-tool calls can't starve unrelated requests (initialize/health) during thread-pool ramp-up.

## Verification

Deployed locally as 5.2.692: `POST /mcp initialize` answers in ~1 ms, no CLOSE_WAIT sockets, external + embedded clients both stable.

### DIFF
--- a/ClarionAssistant/Services/McpServer.cs
+++ b/ClarionAssistant/Services/McpServer.cs
@@ -25,6 +25,11 @@ namespace ClarionAssistant.Services
         private McpToolRegistry _toolRegistry;
         private int _port;
 
+        // Max time a UI-thread MCP tool may run before the request is abandoned
+        // with a timeout error, so a busy/wedged UI thread can't hold a worker
+        // (and leak the connection as CLOSE_WAIT) indefinitely.
+        private const int UiToolTimeoutSeconds = 30;
+
         // Per-session auth token — regenerated on every Start(). Embedded as
         // `Authorization: Bearer <token>` in the MCP config file so the spawned
         // CLI clients (Claude Code, Copilot CLI) transparently authenticate.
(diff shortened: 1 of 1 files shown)


ClarionLive (approved):
Reviewed + verified locally: builds clean, the `BeginInvoke` + bounded-wait replacement for the synchronous `Control.Invoke` correctly kills the permanent-wedge (and the disposed-`ManualResetEventSlim` race after a timeout is properly guarded by the try/catch around `done.Set()`). Two non-blocking notes for posterity:

1. **Doc nit:** the PR description credits fix #2 to `HandleMessagePost`, but the change is (correctly) in `HandleStreamableHttpPost` — the legacy `/messages` handler already closes the response on its early-return paths, and its outer dispatch catch closes on exception.

2. **Known trade-off worth documenting:** delegates abandoned on timeout still sit in the UI thread's queue and will **all execute when it unwedges**. A client that timed out and retried could therefore double-execute a write tool (e.g. an editor mutation). That's the right trade-off versus a permanently dead listener, but a follow-up could set a per-request "abandoned" flag the delegate checks before calling `ExecuteTool`.

Thanks for the thorough field diagnosis — CLOSE_WAIT evidence + root cause made this an easy review.

Issue #30, opened 2026-05-13
ClarionLSP source (LSP server + clarion-indexer) is not publicly available — contributors can't rebuild bundled binaries


bdinko:
## Summary

The repo ships and depends on two compiled artifacts that are not built from source in this repo and whose source doesn't appear to be publicly available anywhere:

1. **`lsp-server/`** — the Node.js Clarion Language Server (`out/server/src/server.js` plus `node_modules/`).
2. **`clarion-indexer.exe`** — the C# CodeGraph indexer that produces `*.codegraph.db` for `query_codegraph`.

Both are pulled from a hard-coded local path on the maintainer's machine at build time:

`installer/ClarionAssistant.iss`:
```
#define SrcClarionIndexer "H:\DevLaptop\ClarionLSP\indexer\bin\Debug"
#define SrcLsp "H:\DevLaptop\ClarionLSP"
```

`installer/build-installer.ps1`:
```
$indexerCsproj = 'H:\DevLaptop\ClarionLSP\indexer\ClarionIndexer.csproj'
...
'H:\DevLaptop\ClarionLSP\indexer\bin\Debug\clarion-indexer.exe',
```

`ClarionAssistant/deploy.ps1`:
```
$IndexerDir    = "H:\DevLaptop\ClarionLSP\indexer"
$IndexerFile   = "$IndexerDir\ClarionIndexer.csproj"
...
$LspSourceDir  = "H:\DevLaptop\ClarionLSP"
```

There is no `H:\DevLaptop\ClarionLSP` repo on GitHub under `peterparker57` or `clarionlive` that I can find, and nothing in `ClarionAssistant` references how to obtain or build these components.

## Why this matters

Anyone who forks the repo to fix a bug or add a feature can rebuild `ClarionAssistant.dll` cleanly with the public source, **but they cannot rebuild `lsp-server` or `clarion-indexer.exe`**. They're forced to copy the pre-built binaries from a released installer back into their deploy output. That works, but:

- The bundled artifacts are pinned to whatever was in the last installer release. If the addin's expectations of the LSP protocol or CodeGraph DB schema ever drift, contributors have no way to keep the components in lockstep.
- Bug fixes inside the LSP server (e.g. the `better-sqlite3` issue below) are impossible to contribute without the source.
- `deploy.ps1` silently skips both components on machines that don't have `H:\DevLaptop\ClarionLSP`, with no documentation on how to obtain or stand them up. New contributors hit this immediately.

[...]

PR #59, opened 2026-07-06
Fix clean-checkout build: PS 5.1 script encoding + indexer SQLite reference

bdinko:
Two fixes that block building/deploying from a fresh clone on **Windows PowerShell 5.1** (the in-box shell) — hit while building v5.2 from source on a machine without PowerShell 7. Both are in the contributor-onboarding spirit of #30/#40.

## 1. `.ps1` scripts break under Windows PowerShell 5.1 (missing BOM)
`deploy.ps1`, `lsp-server-sync/Sync-LspServer.ps1`, and `Update-Version.ps1` are UTF-8 **without a BOM** and contain em-dashes (`—`). Windows PowerShell 5.1 reads a BOM-less `.ps1` as the ANSI code page, so each em-dash (`E2 80 94`) decodes into `â€"` — the stray `"` opens a phantom string and cascades into `missing )` / `missing }` parse errors, and the script won't run at all. PowerShell 7 defaults to UTF-8 so the author never sees it.

**Fix:** add a UTF-8 BOM to the three scripts. Minimal and encoding-preserving; PS7 is unaffected.
_(Alternative if you'd rather avoid a BOM: replace the em-dashes with ASCII `-` so the scripts are pure ASCII. Happy to switch this PR to that approach.)_

## 2. Indexer references an unpublished stub NuGet package
`indexer/ClarionIndexer.csproj` resolves `System.Data.SQLite` via a HintPath into `stub.system.data.sqlite.core.netframework\1.0.118`, which isn't in the NuGet cache and has no `PackageReference` to restore — so the indexer fails to compile on a clean checkout (`CS0234`/`CS0246`).

**Fix:** point the reference and the `CopySQLiteInterop` target at the vendored DLLs the main addin already ships in `ClarionAssistant\lib\sqlite-fts5\` (`System.Data.SQLite.dll` + x86 `SQLite.Interop.dll`), making the indexer self-contained — the actual goal of #30.

## Verification
With both fixes, `deploy.ps1 -Version all` builds the addin (C10/C11/C12) **and** the indexer, and deploys cleanly to all three installs on a PS 5.1 box. No functional/runtime code changed.

### DIFF
--- a/ClarionAssistant/Update-Version.ps1
+++ b/ClarionAssistant/Update-Version.ps1
@@ -1,4 +1,4 @@
-# Update-Version.ps1
+﻿# Update-Version.ps1
 # Pre-build hook called by ClarionAssistant.csproj.
 # 1. Reads Version.props
 # 2. Increments VersionBuild by 1 (unless -NoIncrement is passed)
--- a/ClarionAssistant/deploy.ps1
+++ b/ClarionAssistant/deploy.ps1
@@ -1,4 +1,4 @@
-# ClarionAssistant Deploy Script
+﻿# ClarionAssistant Deploy Script
 # Builds and deploys the addin for Clarion 10, 11, 12, or all.
 # Usage: .\deploy.ps1 [-Version 10|11|12|all] [-NoBuild] [-Kill]
 
--- a/ClarionAssistant/indexer/ClarionIndexer.csproj
+++ b/ClarionAssistant/indexer/ClarionIndexer.csproj
@@ -44,10 +44,11 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
-  <!-- Resolve the NuGet global-packages folder portably (no per-machine user path). -->
+  <!-- SQLite (FTS5): use the vendored DLLs the main addin ships in ..\lib\sqlite-fts5,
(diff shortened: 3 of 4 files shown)


ClarionLive (approved):
Verified locally on both counts:

**Encoding:** reproduced the bug — master's `deploy.ps1` produces **3 parse errors** under Windows PowerShell 5.1 (5.1.26100) exactly as described; with the BOM all three scripts parse clean under both 5.1 and 7.x. Keeping the em-dashes + BOM is the right call over de-ASCII-fying the docs.

**Indexer:** builds clean on the vendored `lib\sqlite-fts5` reference; `CopySQLiteInterop` lands `SQLite.Interop.dll` in `bin\Debug\x86` byte-identical to the lib copy, so the managed/native pair stays matched — and it's the same FTS5 build the main addin already ships, so runtime compatibility is proven.

One follow-up your scan inspired (out of scope here): `installer\build-installer.ps1` and `installer\configure.ps1` are also BOM-less with non-ASCII content, and `configure.ps1` executes on end-user machines under Windows PowerShell during install — today the non-ASCII happens to sit where it doesn't break parsing, but we'll BOM those in a follow-up for consistency.

Thanks for the clean, surgical fix — clean-checkout onboarding matters.

PR #68, opened 2026-07-09
fix(formatter): keyword-named data labels no longer open phantom structures

msarson:
## Problem

Formatting (Ctrl+I) a routine inside an ABC window procedure blows the whole `CODE` body out to the data column (~col 40):

```
SendSignedPing                          ROUTINE
                                        DATA
TestVar                                 LONG
                                        CODE
                                        GLO:gAPIBaseURL = 'https://localhost:7208'
                                        lAccount        = 9999
```

## Root cause

The Smart Formatter keeps a stack of open structures; `ROUTINE`/`DATA`/`CODE` scope transitions only fire when that stack is empty. A column-1 data **label** spelled like a data-structure keyword — most commonly ABC's generated `Toolbar ToolbarClass`, but also `Menu`/`Sheet`/`Tab`/`Option`/`Menubar`/`Form`/`Report` — was read by `structureWord()`/`isDataDecl()` as an **unlabeled structure header**. It opened a `TOOLBAR` structure that never gets an `END`, so the stack stayed open and every following `ROUTINE` and statement cascaded to the data column. It also mis-uppercased the label (`Toolbar` → `TOOLBAR`).

This affects essentially every ABC window procedure that has routines.

Diagnosed by tracing the structure stack over a real 9.2k-line generated `.clw`: at the routine, stack depth was `1`, held open by `line 48: Toolbar ToolbarClass`.

## Fix

A column-1 keyword-spelled token followed by whitespace + another identifier (the type) is a **label**, not a structure header (`Toolbar ToolbarClass`). Genuine unlabeled structures are followed by a comma or nothing (`TOOLBAR,AT(...)`), so they're unaffected; a truly labelled structure whose label matches a keyword (`Group GROUP`) still opens and closes. Restricted to data-structure keywords, so statement/control keywords (`IF X`, `LOOP I`) are untouched.

4-line change across two helpers (`labelLikeStructKeyword` guard added to `structureWord` and `isDataDecl`).

## Tests

`node ClarionAssistant/Terminal/test/clarion-formatter.test.js` → **71 passed, 0 failed** (7 new). New cases: the Toolbar-label cascade (routine body back at one tab), label preserved (not cased), `Toolbar ToolbarClass` opens no structure, `Group GROUP` still balances, real unlabeled `WINDOW`+`MENUBAR` still open/close, and `IF X` not read as a declaration.

### DIFF
--- a/ClarionAssistant/Terminal/clarion-formatter.js
+++ b/ClarionAssistant/Terminal/clarion-formatter.js
@@ -107,12 +107,22 @@
     // The structure/keyword driving a line (uppercased) or ''. If the line begins (column 1) with a
     // non-keyword identifier followed by whitespace, that's a LABEL and the keyword is the next token;
     // otherwise the first token itself is the keyword (handles unlabeled structures like "MENUBAR,USE").
+    // A column-1 token that MATCHES a data-structure keyword but is really a LABEL: it is followed
+    // by whitespace + another identifier (the type), e.g. ABC's "Toolbar ToolbarClass" / "Menu x".
+    // A genuine unlabeled structure header (MENUBAR,USE / TOOLBAR / TOOLBAR,AT()) is followed by a
+    // comma or nothing — never a second identifier — so this stays false for those. Restricted to
+    // DATASTRUCT keywords so statement/control keywords (IF x, LOOP i) are unaffected. Without this,
+    // ABC's "Toolbar ToolbarClass" opens a phantom TOOLBAR structure that never closes, cascading
+    // every following ROUTINE/statement out to the data column (GitHub: keyword-named data labels).
+    function labelLikeStructKeyword(firstTok, code) {
+        return DATASTRUCT_SET[firstTok] && /^[A-Za-z_][A-Za-z0-9_:.]*\s+[A-Za-z_]/.test(code);
+    }
     function structureWord(raw) {
         var code = stripComment(raw);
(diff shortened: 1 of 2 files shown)


ClarionLive:
Reviewed and merged (8c82475) — thanks Mark, excellent diagnosis and fix. The trace-the-structure-stack writeup made this an easy review.

Review notes:
- Your `labelLikeStructKeyword` guard composes cleanly with the MAP-scope work that landed on master this week (5e9480c — MAP is now in `DATA_STRUCT_OPENERS` so prototypes inside a MAP pin to column 1). Your guard actually strengthens it: `Toolbar ToolbarClass`-style labels can no longer open phantom structures that would corrupt the MAP depth tracking, and keyword-named prototypes inside a MAP (`Report PROCEDURE`) now classify as declarations correctly.
- The test-file merge conflict (both branches appended suites at the same anchor) was resolved by keeping both blocks — combined suite is green.
- One follow-up hardening landed on top (955907b): the module-level PREFERRED-keyword branch checked `PREFERRED_SET` without your guard, so a variable named `Map` (`Map MapClass`) still became the MAP keyword, opened a phantom MAP struct, and — with the new mapDepth tracking — dragged everything after it to column 1. Same `labelLikeStructKeyword` guard applied there; 2 more tests. Suite now 80/0.

Ships in the next release.

PR #71, opened 2026-07-09
fix(embeditor): navigate Monaco overlay to native caret position on attach

asantarelli:
## Problem

When the CA Monaco overlay fires via the poll-detect path (ticket 4d16b53a), it was opening Monaco at the last *saved* cursor for that procedure instead of the embed point the developer had the cursor on in the native Clarion embeditor. Net result: click embed point X → Monaco opens at line 1 (or wherever the cursor was last session).

## Root cause

`AttachOverlayToOpenEmbed` never read the native caret position. `SendSource()` only restores from `ModernEmbeditorState` (persisted state), with no override path for the "just-opened from native" case.

## Fix

- `EmbeditorCompletionService.GetNativeCaretLine()` — reads `TextEditorControl.ActiveTextAreaControl.Caret.Line` (0-based → 1-based).
- `ModernEmbeditorLauncher.AttachOverlayToOpenEmbed` — reads the native caret before the deferred `ctx.Post` (the embed is still open at that point) and captures it for the overlay constructor.
- `ModernEmbeditorViewContent` — new `int initialLine = 0` constructor parameter; `SendSource()` overrides `cursorLine` with it once on first paint, then clears it so subsequent navigations are unaffected.

## Test

1. Open a procedure in the native Clarion embeditor (via Clarion's "Embeditor Source" menu).
2. Click on an embed point that is NOT at line 1.
3. The CA Monaco overlay should open with the cursor/view centred on that embed point, not at the top of the file.

### DIFF
--- a/ClarionAssistant/Services/EmbeditorCompletionService.cs
+++ b/ClarionAssistant/Services/EmbeditorCompletionService.cs
@@ -101,6 +101,18 @@ namespace ClarionAssistant.Services
             return true;
         }
 
+        /// <summary>Return the 1-based line the native embeditor caret is on, or 0 on failure.</summary>
+        public static int GetNativeCaretLine()
+        {
+            try
+            {
+                var control = GetActiveEmbeditorControl(new StringBuilder());
+                if (control == null) return 0;
+                return control.ActiveTextAreaControl.Caret.Line + 1; // ICSharpCode caret is 0-based
+            }
+            catch { return 0; }
+        }
+
         /// <summary>
         /// Reflection-based locate of the embeditor's ICSharpCode TextEditorControl,
         /// then a typed cast. Mirrors AppTreeService.GetClaGenEditor()'s search.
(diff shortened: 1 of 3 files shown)


ClarionLive:
Reviewed and merged (8c40c5a) — thanks! Clean, focused fix.

Review notes:
- Caret read timing is right (before the deferred `ctx.Post`, while the native embed is still live, on the trigger's UI-thread stack), and the one-shot `_initialLine` consume is the correct semantic — `OnReady` re-fires on WebView reloads and those now correctly fall back to the persisted cursor.
- Trailing optional ctor param keeps every other call site source-compatible; builds clean against master.
- Non-blocking nit: the override forces `cursorColumn = 1` — reading `Caret.Column` too would restore the exact position, but for landing at an embed point line-only is fine.

Will be live-verified against the poll-detect overlay path and ships in the next release.

ClarionLive:
Follow-up (bbd001d): the legacy **"CA Embeditor" button on the native embeditor toolbar** has been removed from the manifest (`/SoftVelocity/Clarion/ToolBar/EmbedEditor` → `ShowModernEmbeditorCommand`).

That button was the pre-overlay Path-B path (open a Modern mirror view from an open native embed). Since the auto-overlay this PR improves is now the one way the CA Embeditor opens — it attaches itself when a native embed opens — pressing the old button on top of a live overlay opened a second Modern view against the same embed. The native **Open Source** button is unaffected.

Practical effect for this PR's fix: the poll-detect attach path (`AttachOverlayToOpenEmbed`) is now the *only* consumer of the native-caret navigation added here, which is exactly the path it was written for. `ShowModernEmbeditorCommand` remains in the codebase unreferenced (retirement note in the class header) in case a mirror surface is ever wanted again.

PR #74, opened 2026-07-12
feat(embeditor): LSP global scope via real module path + MEMBER header (closes #56)

### LINKED ISSUE #56: Embeditor LSP global scope via synthetic MEMBER header + real module path
Hi John — following the overlay technique you adopted in v5.2, here's another trick from our side ([msarson/ClarionMonacoEditor](https://github.com/msarson/ClarionMonacoEditor)) for the CA Embeditor: getting the **language server itself** to resolve a procedure's **global scope** (global vars, file/field labels, ABC class methods) inside the embed buffer, with no separate index.

**The problem.** The embed buffer is a procedure slice with no module header, addressed by a synthetic filename. The Clarion LSP resolves a member module's global scope by following its `MEMBER('X')` to the program file on disk (roughly `path.resolve(dirOfCurrentFile, X)` + `fs.existsSync`, then reading the parent's global section). A synthetic-named, header-less buffer has neither the real directory nor a `MEMBER`, so program globals never resolve.

**What we do** — two changes when pushing the buffer to the LSP:

1. **Address the buffer by the module's REAL full path.** Read `PweeEditorDetails.AppName` (full `.app` path) + `PweeEditorDetails.Module` -> `dir(AppName)\Module`, the generated `.clw` already on disk and in the project. Gives an exact project match *and* makes the buffer's directory the real project dir.
2. **Prepend the module's own `MEMBER` line**, read verbatim from the on-disk module (it's `MEMBER('App.clw')` — note the `.clw` inside the arg). The server's `MEMBER`->parent resolution then lands on the real PROGRAM `.clw` and pulls in global scope.

One prepended line -> shift editable ranges +1 (and native-caret mirror +1) to keep slots aligned; write-back is unaffected (maps slots to the native doc by index, not Monaco line numbers).

Code: https://github.com/msarson/ClarionMonacoEditor/blob/7f2e1a4c36af579620df509f4119fb8a8ecf6378/ClarionMonacoEditor/MonacoEmbedSecondaryView.cs#L176

**Advantages:** the server resolves scope itself (type-aware, no symbol DB to build/keep fresh); reuses the real generated module + program already on disk; no temp files, no `.cwproj` edits.

**Notes / interplay with your CodeGraph augmentation:**
- Historically the pure upstream LSP only did **member-access** completion (after a `.`) and returned nothing for a **bare prefix** — so on its own this trick didn't cover bare-prefix globals, and your CodeGraph augmentation did. **Heads-up:** bare-prefix completion is now implemented in the upstream LSP **v0.9.9** (finished today, releasing over the next couple of days). Combined with this MEMBER-scope trick, the server itself should then complete bare-prefix globals inside the embed — worth comparing against your CodeGraph path once it's out.
- While the embeditor is open, the buffer **shadows** the on-disk module in the LSP under the same path (could `didClose` on teardown to revert).
- Requires the generated module + program on disk (true after a generate).

Happy to compare notes.

msarson:
## Why

Closes #56.

The CA Embeditor's buffer is a procedure slice with no module header, addressed to the LSP by a **synthetic bare name** (e.g. `file:///UpdateCustomer.clw` — no directory, no file on disk). The Clarion LSP resolves a member module's **global scope** (program globals, file/field labels, ABC class scope) by following its `MEMBER('App.clw')` line to the parent program file on disk — roughly `path.resolve(dirOfCurrentFile, arg)` + `fs.existsSync`. A directory-less, header-less buffer gives that resolution nothing to work from, so PROGRAM-scope symbols never resolved inside the embed. The C#-side CodeGraph merges papered over some of this, but the server itself was blind to global scope.

## What

Two changes when talking to the LSP, per the technique proven in [msarson/ClarionMonacoEditor](https://github.com/msarson/ClarionMonacoEditor) (see #56):

1. **Address the buffer by the generated module's REAL full path** — `PweeEditorDetails.AppName`'s directory + `PweeEditorDetails.Module`. The buffer then lives in the real project directory (exact project match, redirection/libsrc context).
2. **Prepend the module's own `MEMBER(...)` line** — read verbatim from the on-disk module — to every LSP-bound copy of the buffer. MEMBER→parent resolution lands on the real PROGRAM `.clw` and pulls in global scope.

With this, the server itself resolves program globals in embed completion/hover/definition — type-aware, no separate index to build or keep fresh.

## How it fits this codebase (two deliberate adaptations)

- **The MEMBER header is prepended ONLY to the LSP-facing buffer copies — never to the Monaco model.** Save/write-back sends Monaco line numbers straight to `WriteEmbedContentByLine`, and the caret mirror + editable-range decorations all assume a 1:1 line mapping with the native document. So `EmbedLspContext.WrapBuffer` wraps at request time, request positions shift +1 (`LspLine0`), and same-file definition / diagnostic results shift back (`MonacoLine1` / the diagnostics offset). Monaco, ranges, caret and save are untouched. (In ClarionMonacoEditor the header is visible at line 1 because its write-back maps slots by index — that isn't safe here.)
- **The context is captured at mirror time in `ModernEmbeditorLauncher`** (both the snapshot and live-overlay paths), because `PweeEditorDetails` only exists while the native embeditor is open and the snapshot path cancels it before the Monaco tab is constructed.

Details:
- New `Services/EmbedLspContext.cs` — capture (via the existing `AppTreeService.GetOpenPweeDetails()`), verbatim `MEMBER` read with synthesized fallback, buffer wrap, and teardown un-shadow.
- Same-file F12 results: in-buffer ROUTINE targets stay authoritative over the CodeGraph fallback (whose on-disk module line numbers don't map to the assembled buffer).
- Diagnostics landing on the injected header line are dropped before slot-clamping.
- **Un-shadow on close:** while a tab is open its pushed buffer shadows the on-disk module in the server's view under the same path; there's no didClose in the transport, so `Dispose` pushes the on-disk content back (`RevertShadow`).
- Falls back to the classic synthetic-name behavior whenever the context can't be captured (no embed open, module not generated yet) — zero behavior change in that case.

## Testing

- Verified live in the Clarion 11.1 IDE against a real app: PROGRAM globals (e.g. a `VCRRequest LONG,THREAD` from the generated program file) now resolve in embed completion.
- Also verified at the server level with a standalone LSP harness (initialize + `clarion/updatePaths` + didOpen of the MEMBER-wrapped buffer under the real path → completion returns the program global; the same buffer under a synthetic bare name does not).
- Interplay note from #56: upstream LSP v0.9.9+ added bare-prefix completion, so with a bundle re-pin ≥ v0.9.9 the server covers bare-prefix globals in the embed on its own — worth comparing against the CodeGraph merge path afterwards.

## Known edge

Two embeditor tabs open on procedures from the **same generated module** will alternately shadow the same real path in the server (last request wins). Same characteristic as the reference implementation; in practice each request re-pushes its own buffer first, so results stay consistent per request.

### DIFF
--- a/ClarionAssistant/ClarionAssistant.csproj
+++ b/ClarionAssistant/ClarionAssistant.csproj
@@ -140,6 +140,7 @@
     <Compile Include="Services\AppTreeService.cs" />
     <Compile Include="Services\IdeReflectionService.cs" />
     <Compile Include="Services\EmbeditorCompletionService.cs" />
+    <Compile Include="Services\EmbedLspContext.cs" />
     <Compile Include="ShowModernEmbeditorCommand.cs" />
     <Compile Include="ClarionBindingAssemblyResolver.cs" />
     <Compile Include="LspAutostartCommand.cs" />
--- a/ClarionAssistant/Services/EmbedLspContext.cs
+++ b/ClarionAssistant/Services/EmbedLspContext.cs
@@ -1,0 +1,164 @@
+using System;
+using System.IO;
+using System.Reflection;
+
+namespace ClarionAssistant.Services
+{
+    /// <summary>
+    /// Real-module LSP context for the CA Embeditor (GitHub #56 — embed global scope via a synthetic
+    /// MEMBER header + the module's REAL path).
+    ///
+    /// The embed buffer is a procedure slice with no module header, and it used to be addressed to the
+    /// LSP by a synthetic bare name (e.g. "UpdateCustomer.clw", no directory). The Clarion LSP resolves
+    /// a member module's GLOBAL scope (program globals, file/field labels, ABC classes) by following its
+    /// MEMBER('App.clw') to the parent program file on disk — roughly path.resolve(dirOfCurrentFile, arg)
+    /// + fs.existsSync. A directory-less, header-less buffer gives that resolution nothing to work from,
+    /// so program globals never resolved inside the embed.
+    ///
+    /// This context fixes both halves, WITHOUT touching the Monaco buffer itself:
+    ///   1. Address the buffer by the generated module's REAL full path — PweeEditorDetails.AppName's
+    ///      directory + PweeEditorDetails.Module — the generated .clw already on disk and in the project.
+    ///      The buffer then lives in the real project directory (redirection/libsrc/project match).
+    ///   2. Prepend the module's own MEMBER(...) line — read VERBATIM from the on-disk module so the
+    ///      exact argument form matches — to every LSP-bound copy of the buffer. MEMBER→parent resolution
+    ///      then lands on the real PROGRAM .clw and pulls in global scope.
+    ///
+    /// The prepend happens ONLY in the LSP-facing buffer (requests carry <see cref="LineOffset"/>);
+    /// the Monaco model, editable ranges, caret mirror, and save/write-back all keep their existing
+    /// 1:1 line mapping with the native document.
+    ///
+    /// Capture timing matters: PweeEditorDetails exists only while the NATIVE embeditor is open, and the
+    /// snapshot launcher cancels it before the Monaco tab is constructed — so the launcher captures this
+    /// context at mirror time and passes it into the view.
+    ///
+    /// While a context is active, the pushed buffer SHADOWS the on-disk module in the LSP under the same
+    /// path. There is no didClose in the transport, so <see cref="RevertShadow"/> pushes the on-disk
+    /// content back on tab teardown to restore the server to the truth.
+    /// </summary>
+    public sealed class EmbedLspContext
+    {
+        /// <summary>Full path of the generated module .clw on disk (dir of the .app + module name).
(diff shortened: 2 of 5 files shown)


ClarionLive (approved):
Excellent work — the key design decision (MEMBER header prepended ONLY to LSP-bound buffer copies, never the Monaco model) is exactly right for this codebase: save/write-back via WriteEmbedContentByLine, editable-range decorations, and the caret mirror all keep their 1:1 line mapping. Capture-at-mirror-time (PweeEditorDetails dies with the native embeditor) and RevertShadow on dispose show real understanding of the moving parts. The LspLine0/MonacoLine1 conversions are applied consistently across completion, hover, definition, and diagnostics (header-line markers dropped before slot-clamping).

Reviewed as part of the 74→75→78 stack; the full stack builds clean against C12 (5.3.754, no new warnings). One non-blocking note: GetDataSymbols wraps the buffer but doesn't unmap result line numbers — it currently has zero consumers so it's moot, but worth a one-liner if it ever gets wired up.

Merging the stack in order. Thanks Mark!

Issue #72, opened 2026-07-10
Code Completion


bill-atchison:
Do you plan to add code completion to the CA Embeditor?  I find it useful.

ClarionLive:
Good news: the CA Embeditor already has code completion — Ctrl+Space (or just typing) pops Monaco's suggestion list, fed by the bundled Clarion language server (keywords, built-ins, data types, attributes, plus symbols once the server has indexed your solution). Since you're not seeing it, a few things to check, in likelihood order:

1. **Give the language server a beat on cold start.** The first completion request after opening the IDE can come back empty while the LSP is still warming up — try Ctrl+Space again a couple of seconds later. A solution needs to be open for the server to start at all.
2. **Fresh install?** There's a known quirk where Monaco editor features are inert on the very first launch after a clean install — one Clarion restart wires them up. If you installed recently and have never restarted since, try that.
3. **Which CA version are you on?** (Help → About, or the installer filename.) Older builds predate a lot of the current completion pipeline.
4. **Which editor are you actually typing in?** The CA Embeditor is the Monaco surface (dark editor with its own toolbar). If you're in Clarion's classic yellow-tint native embeditor, completion there is more limited.

If none of that gets it working, reply with your CA version + Clarion version and what happens on Ctrl+Space (nothing at all vs. an empty list), and we'll dig in.

Also worth knowing: the next release meaningfully upgrades completion — the bundled language server was just re-pinned to Clarion-Extension v1.0.0 (#77), and #74 makes the embed buffer a real MEMBER module in your app's scope, so completion inside embeds picks up your PROGRAM's globals (GLO:* and prefixed structure fields), with signature help from #75 on top.

PR #75, opened 2026-07-12
feat(editor): signature help (parameter hints) in the Monaco embeditor + source editor

msarson:
> **Builds on #74** — the branch is stacked on `feat/embeditor-member-scope`, so this diff shows those commits too until #74 merges. This PR's own changes are the last two commits (`5f43fb4`, `7a8e08b`).

## What

Typing `(` (retrigger on `,`) at a call site now shows the VS Code-style parameter-hint overlay — signature label, active parameter bolded, and the `1/4` overload pager — in both the CA Embeditor and the Monaco source editor. Fed by `textDocument/signatureHelp`, which the Clarion LSP has advertised all along but nothing in the addin consumed.

Related: #72 — together with #74 this rounds out the embeditor completion experience (PROGRAM-scope completion from #74, parameter hints from this PR).

## How

- **`monaco-embeditor.html`**: `registerSignatureHelpProvider` (`(` trigger, `,` retrigger), suppressed inside comments like hover. One registration serves both surfaces (same page).
- **`MonacoEditorControl`**: new `{action:"signatureHelp"}` host callback on `IMonacoEditorHost`.
- **`ModernEmbeditorViewContent`**: handler with the same position/buffer contract as hover — including the #74 real-module wrap, so hints resolve with full PROGRAM scope inside embeds.
- **`MonacoClarionSourceEditor`**: same handler against the real file path.
- **`LspClient.GetSignatureHelp`**: buffer-aware `textDocument/signatureHelp` (bundled server ≥ v0.9.6 advertises the capability).
- **`SharedLspBridge.GetSignatureHelp`**: bundled path parses the raw LSP shapes (string-or-MarkupContent docs, string-or-`[start,end]`-offset parameter labels). The SHARED path invokes `GetSignatureHelpAsync` via **reflection**: the method shipped in ClarionLsp contracts **v1.4.0**, newer than the vendored `ClarionLsp.Contracts.dll` we compile against — and per the class-header JIT-safety note (both assemblies are AssemblyVersion 1.0.0.0, first-load-wins), a static reference wouldn't be safe anyway. A ClarionLsp < 1.4.0 install logs once and returns null; the widget simply stays hidden.

## The bug worth knowing about (second commit)

First cut rendered the label + arrows but a blank overload counter. Root cause: the bridge emitted `"documentation": null` on every signature dict. Monaco 0.52's render gate is `documentation !== undefined`, so an explicit null enters the markdown-docs path and aborts the render midway — after the label paints, before the counter text is written (`.overloads` left `""`/`visibility:hidden`/`width:0`). Reproduced deterministically outside the IDE by driving the real `monaco-embeditor.html` in headless Edge with a shimmed `chrome.webview` host: same payload with `documentation: null` → blank counter; key omitted → `1/4` renders. Fix: `SigDict()` only emits `documentation` when non-empty, on both paths. If you ever hand Monaco widget data from C#, omit empty keys rather than passing nulls.

## Testing

- Live in Clarion 11.1: `Body.SetValue(` (StringTheory) in a code embed shows all 4 overloads with the pager; `↑`/`↓` cycle; active parameter bolds across commas. Verified on the shared-addin path (ClarionLsp 1.4.0).
- Server-level via a standalone LSP harness (initialize + `clarion/updatePaths` + didOpen of the MEMBER-wrapped buffer): `WebClient.Fetch(` returns both overloads (`String` / `StringTheory`) with parameters and active indices.
- Widget-level via the headless-Edge harness above, light and dark themes.

### DIFF
--- a/ClarionAssistant/ClarionAssistant.csproj
+++ b/ClarionAssistant/ClarionAssistant.csproj
@@ -140,6 +140,7 @@
     <Compile Include="Services\AppTreeService.cs" />
     <Compile Include="Services\IdeReflectionService.cs" />
     <Compile Include="Services\EmbeditorCompletionService.cs" />
+    <Compile Include="Services\EmbedLspContext.cs" />
     <Compile Include="ShowModernEmbeditorCommand.cs" />
     <Compile Include="ClarionBindingAssemblyResolver.cs" />
     <Compile Include="LspAutostartCommand.cs" />
--- a/ClarionAssistant/MonacoClarionSourceEditor.cs
+++ b/ClarionAssistant/MonacoClarionSourceEditor.cs
@@ -483,6 +483,26 @@ namespace ClarionAssistant
             });
         }
 
+        // LSP signature help — parameter hints when typing '(' or ',' at a call site. Same
+        // position/buffer contract as OnHover; the page hides the widget on a null reply.
+        void IMonacoEditorHost.OnSignatureHelp(MonacoEditorControl editor, string rawJson)
+        {
+            int reqId, line, col; string buffer;
+            if (!ParseLspRequest(rawJson, out reqId, out line, out col, out buffer)) return;
+            System.Threading.Tasks.Task.Run(() =>
+            {
+                Dictionary<string, object> help = null;
+                try
+                {
+                    EnsureLsp();
+                    if (SharedLspBridge.IsRunning)
+                        help = SharedLspBridge.GetSignatureHelp(_filePath, Math.Max(0, line - 1), Math.Max(0, col - 1), buffer);
+                }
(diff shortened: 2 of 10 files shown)


ClarionLive (approved):
Approved as part of the 74→75→78 stack review (full stack builds clean against C12). Two things worth calling out: the reflection path for GetSignatureHelpAsync (contracts v1.4.0 is newer than the vendored copy we compile against) with graceful degradation + a clear log line on older addins is exactly the pattern this repo's strong-name history demands; and the Monaco 0.52 null-documentation render-abort fix (omit the key vs explicit null) is a genuinely subtle catch — the comment explaining it will save someone hours.

PR #78, opened 2026-07-12
feat(editor): Ctrl+F12 go-to-implementation in the Monaco embeditor + source editor

msarson:
> **Builds on #75** (which builds on #74) — the branch is stacked, so this diff shows those commits until they merge. This PR's own changes are the last commits from `78ac4b5` onward.

## What

**Ctrl+F12** jumps to a symbol's **implementation body** via `textDocument/implementation` — from a declaration OR from any call site/reference. It's the "skip the prototype" navigation: where plain F12 on a class method typically lands on the `.inc` prototype, Ctrl+F12 lands on the executable body. A capability the Clarion LSP has advertised since before the current bundle pin, with nothing consuming it. Works in both the CA Embeditor and the Monaco source editor:

- **From a call site** (the everyday case): Ctrl+F12 on `Body.SetValue(` (Body is a StringTheory) opens **`StringTheory.clw` at the SetValue implementation** — plain F12 on the same call goes to the `.inc` prototype.
- **From a declaration**: Ctrl+F12 on `Init` in `ThisWindow CLASS(WindowManager)` jumps to the `ThisWindow.Init PROCEDURE` body further down the embed buffer.
- In the source editor: Ctrl+F12 in an `.inc` prototype jumps to the `.clw` body.

## How

- `monaco-embeditor.html`: Ctrl+F12 branch in the capture-phase key interceptor, ahead of plain F12 (which is untouched); seeds Navigate Back the same way.
- `MonacoEditorControl`: `{action:"implementation"}` host callback.
- Both hosts: handlers mirroring their definition flows — same-file targets reveal in-place (unwrapping the #74 MEMBER-header offset in embed mode), cross-file targets open via `MonacoSourceNavigator`.
- `SharedLspBridge.GetImplementation`: SHARED path invokes the addin's `GetImplementationAsync` via **reflection** (it entered the ClarionLsp contracts after the vendored copy — same pattern and rationale as #75) and flattens `LocationResult[]` into the raw LSP location dicts `TryGetFirstLocation` already understands. The contract method takes no `bufferText`, so the live buffer is synced explicitly first. An older addin logs once and the keypress no-ops. BUNDLED path is a raw `textDocument/implementation` request.

## The bug worth knowing about

First cut: the addin resolved the location (`GetImplementation result: 1 location(s)`) but nothing opened. Root cause: the server percent-encodes the drive colon per VS Code URI conventions (`file:///c%3A/Clarion/.../StringTheory.clw`), and both `UriToFilePath` implementations (SharedLspBridge + bundled LspClient) only decoded `%20` — producing `c%3A\Clarion\...` paths that fail `File.Exists`, which `NavigateToFileAndLine` treats as a silent false **before any logging**.

F12 masked this for years because `GetDefinition` preempts the LSP on library member access with the ClarionGraph DB lookup (clean Windows paths) — implementation is pure-LSP and was the first feature to walk the encoded path. Both converters now run `Uri.UnescapeDataString` (with a minimal `%3A`/`%20` fallback), which also heals the same `%20`-only gap in the shared ClarionLsp addin's own `LocationResult.FilePath` values via the re-encode/decode round-trip (tracked separately in msarson/clarion-lsp#3). Any pure-LSP location consumer benefits, not just Ctrl+F12.

## Testing

- Live in Clarion 11.1 (shared ClarionLsp 1.4.0 path): Ctrl+F12 on the `Body.SetValue(` call site opens `StringTheory.clw` at the implementation; Ctrl+F12 on a CLASS-block method declaration reveals the body in-buffer; plain F12 behavior unchanged.
- Standalone server harness: `textDocument/implementation` on the `ThisWindow.Init` declaration resolves to the `ThisWindow.Init PROCEDURE` body line; from the `Body.SetValue(` call site resolves to `file:///c%3A/.../StringTheory.clw` — the URI shape that exposed the decode bug.

### DIFF
--- a/ClarionAssistant/MonacoClarionSourceEditor.cs
+++ b/ClarionAssistant/MonacoClarionSourceEditor.cs
@@ -483,6 +483,39 @@ namespace ClarionAssistant
             });
         }
 
+        // Ctrl+F12 go-to-implementation — declaration → implementation body. Same navigation shape
+        // as OnDefinition: same-file targets reveal in-place, cross-file targets open the file.
+        void IMonacoEditorHost.OnImplementation(MonacoEditorControl editor, string rawJson)
+        {
+            int reqId, line, col; string buffer;
+            if (!ParseLspRequest(rawJson, out reqId, out line, out col, out buffer)) return;
+            System.Threading.Tasks.Task.Run(() =>
+            {
+                bool navigated = false;
+                try
+                {
+                    EnsureLsp();
+                    if (SharedLspBridge.IsRunning)
+                    {
+                        var impl = SharedLspBridge.GetImplementation(_filePath, Math.Max(0, line - 1), Math.Max(0, col - 1), buffer);
+                        string targetPath; int targetLine0, targetChar0;
+                        if (SharedLspBridge.TryGetFirstLocation(impl, out targetPath, out targetLine0, out targetChar0))
+                        {
+                            if (SameSourceFile(targetPath))
+                            {
+                                if (_editor != null) _editor.RevealLine(targetLine0 + 1, targetChar0 + 1);
+                                navigated = _editor != null;
+                            }
+                            else
+                                navigated = MonacoSourceNavigator.NavigateToFileAndLine(targetPath, targetLine0 + 1, 1);
+                        }
+                    }
+                }
+                catch (Exception ex) { MonacoSpikeLog.Write("overlay implementation error: " + ex.Message); }
+                editor.PostResponse(reqId, new Dictionary<string, object> { { "navigated", navigated } });
+            });
+        }
+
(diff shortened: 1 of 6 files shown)


ClarionLive (approved):
Approved — completes the 74→75→78 stack (full stack build-verified against C12). The URI percent-decoding fix riding along here is a real catch: the server encodes the drive colon (file:///c%3A/...) and the old %20-only decode made every pure-LSP navigation silently no-op — fixed consistently in both LspClient and SharedLspBridge. The reflection path for GetImplementationAsync degrades gracefully on older ClarionLsp addins, same as signature help.

Note: I rebased this branch onto master after squash-merging #74/#75 — the stack's shared commits are already on master, so the branch now carries only the two go-to-implementation commits (content unchanged).

PR #91, opened 2026-07-14
Support VS2026 Build Tools, add Clarion 11.1 as a distinct target, make installer paths portable

OkayPlunk:
## Why

Building on a dev machine with **Visual Studio 2026 Build Tools** (no VS2022 IDE, no Inno Setup) instead of a machine set up like the original author's failed immediately: `build-installer.ps1` hardcoded VS2022 MSBuild paths and a fixed Inno Setup location, and `ClarionAssistant.iss` hardcoded absolute paths onto one specific developer's machine (`H:\DevLaptop\...`, `C:\Users\John Hickey\...`). Along the way, testing also surfaced that Clarion 11 and 11.1 were being treated as the same target even though they're separate releases with separate binding DLLs — risking an ABI mismatch if both are installed on one machine.

## What changed

**`build-installer.ps1`**
- Locates MSBuild across VS2022/2026, Community/Professional/Enterprise/Build Tools, with a `vswhere` fallback for future versions.
- Locates Inno Setup 6 at its per-user install path (`%LOCALAPPDATA%\Programs\...`) in addition to the system-wide path.
- The per-config version-freshness gate now also watches the `C11.1` bin, alongside C10/C11/C12.
- Stopped forcing `/p:Platform=AnyCPU` on the `ClarionIndexer` build. That project only defines `<OutputPath>` under its `Debug|x86`/`Release|x86` conditions (x86 is its default when `Platform` is unset) — forcing `AnyCPU` matched neither, so MSBuild failed with "BaseOutputPath/OutputPath property is not set" every time. `deploy.ps1` already built it correctly (no `Platform` override); this now matches.
- `docgraph.db` is now auto-copied from its real default location (`%APPDATA%\ClarionAssistant\docgraph.db`, i.e. `DocGraphService.GetDefaultDbPath()` in the C# runtime) into the installer folder when missing there, instead of just warning and requiring a manual copy every release.

**`ClarionAssistant/deploy.ps1`**
- Resolves each Clarion version's install root via the `SoftVelocity` registry keys first (authoritative on modern Clarion installs), falling back to known hardcoded paths, then a drive-root glob scan (`Clarion12*`, etc.) as a last resort.
- Splits Clarion 11 and 11.1 into **distinct** build/deploy targets — confirmed via the registry that they're separate installs (`SoftVelocity\Clarion11` vs `SoftVelocity\Clarion11.1`/`Clarion111`), not aliases of each other.
- A missing Clarion version is now a clean skip instead of aborting the whole run (previously one missing version blocked every later version from building).
- Removed a fallback in `Resolve-BuildOutputDir` that could silently deploy a build compiled against one Clarion version's binding DLLs into a *different* version's live install when the correct per-version output folder didn't exist yet. Caught this via an actual reproduction during testing (a Clarion-12-built DLL got copied into a Clarion 11.1 install) before it shipped.
- `vswhere` is now called with `-products *` so it also finds Build Tools-only installs (by default it only sees Community/Professional/Enterprise).

**`Directory.Build.props`**
- Split the `ClarionVersion` `11`/`11.1` `ClarionRoot` defaults to match the above, each probing a couple of known install paths rather than one hardcoded value.

**`installer/ClarionAssistant.iss`**
- Replaced hardcoded absolute paths with paths resolved relative to the script's own location (`{#SourcePath}`) or via environment variables (`%USERPROFILE%`, `%APPDATA%`, plus overrides like `CLARIONCOMBROWSER_DIR`, `CLARION12_ROOT` for repos that live outside this one).
- External dependencies that aren't part of this repo (ComForClarion, UltimateCOM templates/classes, ClarionCOM tooling, bundled Node.js) are now wrapped in `#ifexist` guards — same pattern already used for `docgraph.db` — so the installer compiles cleanly on a machine that doesn't have those side-repos checked out, instead of failing to find a file.
- Added a full **Clarion 11.1** component: file list, wizard page field + Browse button, path validation, component auto-check, and post-install cleanup — mirroring 10/11/12. The wizard's own Clarion-path auto-detection (`DetectClarionPaths`) now also checks 11 vs 11.1 as separate registry keys, matching the `deploy.ps1` fix.
- Fixed a Local vs Roaming AppData mismatch: the `docgraph.db` `[Files]` entry installs to `{userappdata}` (Roaming, matching `DocGraphService.GetDefaultDbPath()`'s real `Environment.SpecialFolder.ApplicationData` path), but the `-DocGraphDb` parameter passed to `configure.ps1` pointed at `{localappdata}` (Local) — a folder the file was never actually placed in. Fixed both the `[Run]` parameter and the matching `[Dirs]` entry. (Currently latent/harmless — nothing in the C# code reads the `CLARION_DOCGRAPH_DB` env var `configure.ps1` sets, `DocGraphService` always calls its own `GetDefaultDbPath()` — but worth being correct in case that wiring gets added later.)

## Testing

Verified end-to-end on a machine with VS2026 Build Tools and Clarion 10, 11, 11.1, and 12 all installed:
- `build-installer.ps1` builds all four addin configs and packages the installer.
- Ran the compiled installer for real (elevated, silent) — it installed correctly into `C:\Clarion10\accessory\addins\ClarionAssistant`.
- Launched the actual Clarion 10 IDE afterward: the addin's autostart log fired and succeeded within the same second the IDE started, and the "Clarion Assistant" entry appeared in the Tools menu.
- `ClarionIndexer` now builds cleanly to `bin\Debug\clarion-indexer.exe` with the corrected command; re-compiled `ClarionAssistant.iss` successfully after all changes above.

## Known pre-existing gap not addressed here

Flagging for visibility, not fixed in this PR: the per-config version-freshness gate in `build-installer.ps1` can't fully agree with itself when `deploy.ps1` builds multiple configs in one pass, because `VersionBuild` auto-increments on every individual MSBuild compile — only the last-built config matches by the time packaging runs.

### DIFF
--- a/ClarionAssistant/deploy.ps1
+++ b/ClarionAssistant/deploy.ps1
@@ -1,9 +1,9 @@
 ﻿# ClarionAssistant Deploy Script
-# Builds and deploys the addin for Clarion 10, 11, 12, or all.
-# Usage: .\deploy.ps1 [-Version 10|11|12|all] [-NoBuild] [-Kill]
+# Builds and deploys the addin for Clarion 10, 11, 11.1, 12, or all.
+# Usage: .\deploy.ps1 [-Version 10|11|11.1|12|all] [-NoBuild] [-Kill]
 
 param(
-    [ValidateSet("10","11","12","all")]
+    [ValidateSet("10","11","11.1","12","all")]
     [string]$Version = "all",  # Which Clarion version(s) to build/deploy
     [switch]$NoBuild,          # Skip build, just copy
     [switch]$Kill              # Kill Clarion IDE before deploying
@@ -16,7 +16,7 @@ $ErrorActionPreference = "Stop"
 function Resolve-MSBuild {
     $vswhere = Join-Path ${env:ProgramFiles(x86)} "Microsoft Visual Studio\Installer\vswhere.exe"
     if (Test-Path $vswhere) {
-        $found = & $vswhere -latest -requires Microsoft.Component.MSBuild `
+        $found = & $vswhere -latest -products * -requires Microsoft.Component.MSBuild `
                             -find "MSBuild\**\Bin\MSBuild.exe" | Select-Object -First 1
         if ($found -and (Test-Path $found)) { return $found }
     }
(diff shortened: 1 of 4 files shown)


ClarionLive (approved):
Thanks Paul — this is excellent work, and exactly the kind of contribution this repo needs. The registry-first root resolution, the VS2026/Build Tools support, and especially the removal of the `Resolve-BuildOutputDir` fallback (which really could deploy a DLL built against one Clarion version's binding DLLs into a *different* version's live install — and you caught it with an actual reproduction) are all genuinely valuable. The 11 / 11.1 split is correct and correctly reasoned, and the comment quality throughout matches this repo's conventions perfectly.

While reviewing we found one subtle but serious issue, and rather than bounce the PR back we pushed a fix commit directly to your branch (you had "allow edits by maintainers" enabled — thank you): **efd4b50**.

## The issue: ISPP `#ifexist` matches FILES only

`#ifexist` (and `FileExists`) return **FALSE for directories**, even when the directory exists. Verified empirically against Inno Setup 6's ISCC:

```
#ifexist "C:\Windows"             → FALSE   (directory!)
#ifexist "C:\Windows\notepad.exe" → TRUE
#if DirExists("C:\Windows")       → TRUE
```

Most of the new guards targeted directories (`SrcLsp + "\out\server"`, `SrcComForClarion`, `SrcUltimateClasses`, `SrcBlankDct`, `SrcAgents`, ...), so those `[Files]` blocks were being silently excluded from every compiled installer — including the **entire bundled LSP server** for all four Clarion versions, ComForClarion, the UltimateCOM templates, blank.dct + ClassModels, and the Claude quality agents. Your smoke test (addin loads, Tools menu appears) couldn't catch it because the addin runs fine without the bundled LSP via its shared-LSP fallback.

## What efd4b50 changes

- All guards converted to `Have*` presence flags defined in one place, each probing a **sentinel file** (e.g. `HaveLsp` → `out\server\src\server.js`, `HaveBlankDct` → `blank.dct`); only the wildcard-only ComDocs tree uses `DirExists`.
- Each missing source now emits one loud `#pragma message "WARNING: ... shipping WITHOUT ..."` — silent omission is what made this bug invisible, so the packaging log now always shows what was left out.
- The whole **Clarion 11.1 `[Files]` block** is now guarded on `HaveC11_1`. It was unguarded, so ISCC hard-failed on any machine without `bin\Debug-C11.1` — which contradicted `build-installer.ps1`'s own freshness gate that treats a missing config bin as "won't ship this config". Now it warns and skips, consistent with the gate.
- `deploy.ps1`: a version whose build output folder is missing is now a clean skip. Without this, `-NoBuild` on a never-built config would create the live `accessory\addins\ClarionAssistant` folder and fill it with indexer/LSP/SQLite files but **no addin DLL** — the same "no guessing" principle as your `Resolve-BuildOutputDir` fix, applied one level up.

## Verification

Full ISCC compile on a machine with `Debug-C10/C11/C12` bins, the `.lsp-build\v0.9.8` LSP build, and **no** `Debug-C11.1`:

- Successful compile (89s, 980 files).
- LSP `server.js` ×3 + `node.exe`, blank.dct, agents, ClarionCOMBrowser.dll, UltimateCOM.inc, UCSelectCOM.dll, and all three per-version addin DLLs confirmed **in** the package — all of which were missing before the fix.
- Exactly one warning emitted: the intended C11.1 skip. No hard failure.

One note for anyone building releases from this: the numbered-release recipe becomes a **four-config** build (C10/C11/C11.1/C12) on machines that have all four Clarion versions installed — the freshness gate already watches all four thanks to this PR.

With efd4b50 on the branch this is ready to merge. Thanks again — and the "Known pre-existing gaps" section in your PR description was genuinely helpful for review.

OkayPlunk:
Thank you for catching this, and for pushing the fix directly rather than bouncing the PR back — genuinely a serious bug, and I appreciate you taking the time to explain it this clearly.

I didn't take the diagnosis on faith — reproduced it independently before agreeing:

```
#ifexist "C:\Windows"             -> FALSE   (directory)
#if DirExists("C:\Windows")       -> TRUE
#if FileExists("C:\Windows")      -> FALSE
#ifexist "C:\Windows\notepad.exe" -> TRUE
```

Matches exactly. My directory-target `#ifexist` guards (`SrcLsp + "\out\server"`, `SrcComForClarion`, `SrcUltimateClasses`, `SrcBlankDct`, `SrcAgents`) were unconditionally false, so those `[Files]` blocks never shipped on *any* machine, not just ones missing the sources. My own smoke test (addin loads, Tools menu appears) couldn't have caught this since the addin degrades gracefully without the bundled LSP — exactly as you noted.

Also compiled `efd4b50` on my own machine (which has C10/C11/C11.1/C12 built, but is missing node.exe/LSP/ComForClarion/UltimateCOM/ClarionCOM-tooling/blank.dct/agents, same as your test machine minus the C11.1 gap):

- Successful compile, one loud `#pragma` warning per missing source — all 8 expected warnings fired, nothing silent.
- `HaveC11_1` correctly resolved true here (since `bin\Debug-C11.1\ClarionAssistant.dll` exists on this machine) and included the whole C11.1 block, confirming that guard's condition works both ways, not just the false/skip path you tested.

Two things I specifically like about the fix beyond just correcting the bug: the sentinel-file probes are a real improvement over even a plain `DirExists()` swap — they check that the *expected content* is actually there (e.g. `ClarionAssistant.dll` for C11.1, `server.js` for the LSP build), not just that a directory happens to exist, which would also catch a half-built output folder. And keeping `DirExists()` only for `SrcComDocs` is the right call — it's the one tree copied with a bare wildcard and no stable filename to probe, so a sentinel file wouldn't apply there anyway.

No notes or concerns from my side — this is correct and well-verified. Thanks again.

Issue #90, opened 2026-07-14
Codegraph: Procedure-local derived class misclassified as global - and its inline override corrupts call attribution in a separate scan


geircodes:
## Summary

Investigating the same real-world procedure that turned up a previous local-variable gap, another local variable was missing — but this time the effect was more surprising than "absent." A `CLASS` declared inside a procedure's own `DATA` section, deriving from another class and overriding one of its virtual methods inline (a real, documented pattern for customizing certain library classes' parsing behavior), showed up as a **top-level global class** instead — and every call made anywhere inside that procedure, for the rest of its body, got silently attributed to a completely unrelated procedure elsewhere in the same file.

This turned out to be two independent bugs stacked together, in two different parsing passes that don't share any state. The first was caught by inspecting symbols. The second was only caught by inspecting relationships — after the first fix looked complete and correct at the symbol level.

---

## Root cause

**Bug 1 — symbol misclassification.** The parser that extracts symbols from a procedure implementation has a check that recognizes `Name CLASS(ParentType)` lines and treats them as a new top-level class declaration — entering "scan for method prototypes" mode and recording the name as a global `class` symbol. That check fires unconditionally, with no regard for whether the parser is currently inside a procedure's own `DATA` section. A class declared *inline*, as a local variable of that procedure, gets exactly the same treatment as a genuine file-level class declaration — so it's recorded as a phantom global class instead of a local variable, and it's completely absent from the procedure's own local-variable list.

**Bug 2 — call misattribution, found only after fixing Bug 1.** There is a second, entirely separate parsing pass, used only to detect call sites and figure out which procedure "owns" each one as it scans through a file line by line. This pass shares no state at all with the symbol-extraction pass — it's an independent re-implementation of similar logic, for a different purpose. It has no concept of a local class body either.

[...]

Issue #88, opened 2026-07-14
CodeGraph: Self-closing inline GROUP(Type) END/. in .inc CLASS body leaks classEndDepth, silently deletes every class declared afterward in the file


geircodes:
## Summary

Asking CodeGraph "who uses this class's two specific data members" turned up nothing — not just for those two members, but for **every single data member of the class**. Its methods indexed fine; every data member was invisible. Digging in found a single-line class property, `SomeMember GROUP(SomeGroupType) END` — a named `GROUP,TYPE` instantiated inline, all on one line, and a genuinely common Clarion pattern — sitting near the top of the class. That one line was enough to permanently wedge the `.inc` parser's internal state, and the damage didn't stop at the end of that class: **every class declared afterward in the same `.inc` file silently stopped existing in the index too**, cascading rather than resetting.

---

## Root cause

`ParseIncFile`'s "nested GROUP/QUEUE/RECORD" check exists to track when a `CLASS` body contains an inner structure, so its own `END` doesn't get mistaken for the end of the class. It matches any line shaped like `Name (GROUP|QUEUE|RECORD)...` and increments a depth counter, `classEndDepth`, on the assumption that a separate, later line consisting of just `END` will come along to decrement it back down.

That assumption breaks for the *self-closing* single-line form — `InlineGroup GROUP(SmallGroupType) END` opens and closes on the very same line. Nothing ever decrements the counter for it, because the "is this line just a bare `END`" check requires the terminator to be the *only* thing on the line — which it isn't here, since `InlineGroup GROUP(SmallGroupType)` comes before it. The counter leaks upward by one, permanently.

Two compounding consequences, in increasing order of severity:

1. Capturing a class's data members is gated on the depth counter reading exactly 1 (i.e., "directly inside the class body, not inside some nested structure"). Once it leaks to 2, every member declared afterward in that class is silently skipped — nothing is added, no error, no trace.
2. Far worse: the check that recognizes the *start of a new class* only ever runs when the parser believes it isn't currently inside a class body. Since the leaked counter never lets that flag clear, the parser behaves as if it's *still* inside the first class, forever — meaning **every class declared later in the same file is never even recognized as a class**, let alone have its members captured.

Confirmed directly: after one such self-closing line, two classes declared later in the same file both vanished as `class` symbols entirely — not just missing members, missing *entirely*, even though neither had anything wrong with its own declaration.

[...]

Issue #79, opened 2026-07-13
CodeGraph drops repeat calls to the same target from a procedure


geircodes:
## Summary

`ResolveRelationships` silently drops every call/reference relationship after the first one when the *same* caller invokes the *same* target more than once. Concretely: if procedure `Foo` calls `Worker.Sign(...)` on two different lines, only the first call site is ever written to the `relationships` table — the second (and any subsequent) call is treated as a "duplicate" and discarded, even though it is a distinct, real call site with its own line number.

Found while re-investigating `who calls WorkerClass.Sign` (see #54) in a real production solution. After fixing the two bugs from #54, one call site was still missing: a `worker.Sign(...)` call inside a test procedure that *also* calls `worker.Sign(...)` earlier in the same procedure, on a different line. The first call site was captured; the second was not.

---

## Root cause

The de-duplication key used to avoid inserting the same relationship twice does not include the line number:

```csharp
// Track inserted relationships to avoid duplicates: "fromId|toId|line"
var insertedRels = new HashSet<string>();
...
string relKey = string.Format("{0}|{1}|calls", currentProcId, targetId);
if (insertedRels.Add(relKey))
{
    _db.InsertRelationship(new ClarionRelationship
    {
        FromId = currentProcId,
        ToId = targetId,
        Type = "calls",
        FilePath = file.FullPath,
        LineNumber = i + 1
    });
    relCount++;
}
```

The inline comment claims the key is `"fromId|toId|line"`, but the actual format string is `"fromId|toId|calls"` — line number was never part of it. The moment procedure `Foo` (id 100) calls target `Worker.Sign` (id 200) once, `insertedRels` contains `"100|200|calls"` permanently for the rest of that file's scan. Any later call from the same procedure to the same target — on a different line — hits `insertedRels.Add(relKey) == false` and is skipped. The row that *does* get inserted still carries a correct `LineNumber` (whichever call site happened to be seen first), which makes the bug easy to miss: the relationship isn't wrong, it's just incomplete — one row instead of N.

This pattern appears at 5 call sites in `ResolveRelationships`, all following the same shape:

| Site | Relationship type | Trigger |
|---|---|---|
| `START(ProcName, ...)` detection | `calls` | Thread start |
| `SELF.Method` / `PARENT.Method` detection | `calls` | Class method dispatch |
| `ObjectName.MethodName(...)` detection | `calls` | Typed-variable method call (the `Worker.Sign` case) |
[...]

PR #92, opened 2026-07-15
CodeGraph: CLASS-body member capture in .inc files misses LIKE(...) and EQUATE-aliased-type members

geircodes:
# CLASS-body member capture in `.inc` files misses `LIKE(...)` and `EQUATE`-aliased-type members

## Summary

While re-verifying a previous fix (#88) against a real production class, one of its data members was still conspicuously missing — even though every other member on the same class, including several right next to it, now indexed correctly. The missing member was declared with `LIKE(SomeOtherType)`, borrowing its layout from another structure entirely, rather than being a plain scalar field or a `&Reference`.

Digging further turned up a second, closely related gap in the same place: a member typed via a custom, `EQUATE`-aliased scalar-type synonym (a common pattern for giving a plain numeric type a more descriptive name, e.g. a handle or context type) was equally invisible, for the exact same underlying reason.

## Root cause

`ParseIncFile`'s CLASS-body member-capture cascade only ever tries two declaration patterns against each member line: one for a plain scalar type, one for a `&Reference`. `ParseMemberFile`'s own `DATA`-section cascade tries six, for the identical purpose.

Neither `Name LIKE(OtherType)` nor `Name MyAliasedType` (where `MyAliasedType EQUATE(LONG)`) fits either of the two patterns tried — both simply match nothing, and the member silently never becomes a symbol.

## Fix

Extended the cascade to also try `LikeDeclRegex`, `EquateDeclRegex`, and (last, broadest) `ClassInstanceDeclRegex` — mirroring `ParseMemberFile`'s own ordering. `GroupQueueDeclRegex` is deliberately **not** added: a `GROUP`/`QUEUE`/`RECORD`-shaped member line is already intercepted earlier by the existing depth-tracking check from #88, which manages `classEndDepth` but doesn't create a symbol for the member's own name — a separate, narrower, still-open gap.

## Verification

Extended `test-fixtures/codegraph-repro` with `LikeMemberBugClass` (`GenCertData LIKE(SmallGroupType)`, `SomeHandle` typed via a new `SmallHandleType EQUATE(LONG)` synonym). Both now correctly appear as symbols — 18 total callers of `WorkerClass.Sign`, up from 17. No regressions on any prior fix (#79–#90).

Also verified against a real production solution: the `LIKE(...)`-declared member that surfaced this gap now correctly appears, and its owning class shows its complete member list.

### DIFF
--- a/ClarionAssistant/CodeGraph/Parsing/ClarionParser.cs
+++ b/ClarionAssistant/CodeGraph/Parsing/ClarionParser.cs
@@ -965,32 +965,92 @@ namespace ClarionCodeGraph.Parsing
                         // Strip a trailing inline comment before matching -- see the DATA-section
                         // fix above for why (e.g. "PrivKey &SomeClass !some comment" must still match).
                         string memberForTypeMatch = StripInlineComment(trimmedMember);
+
+                        // Only RefVariableDeclRegex/VariableDeclRegex were ever tried here, unlike
+                        // ParseMemberFile's own DATA-section cascade, which tries six different
+                        // declaration shapes for the identical purpose. A member declared via
+                        // LIKE(OtherType) (e.g. "GenCertData LIKE(GenCertGroupType)", borrowing
+                        // another structure's layout) or typed via a custom EQUATE-aliased scalar
+                        // synonym (e.g. "CRYPT_CONTEXT EQUATE(LONG)") matched neither regex and
+                        // silently never became a symbol at all (issue: LIKE()/EQUATE-alias-typed
+                        // CLASS member never captured). GroupQueueDeclRegex is deliberately NOT
+                        // added here: a GROUP/QUEUE/RECORD-shaped member line is already
+                        // intercepted earlier, by the "Nested END for inner GROUP/QUEUE etc."
+                        // check above, so it can never reach this cascade at all.
+                        string memberName = null;
+                        string memberType = null;
+                        string memberAttrs = "";
+
                         var refMatch = RefVariableDeclRegex.Match(memberForTypeMatch);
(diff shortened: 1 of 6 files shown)


ClarionLive (approved):
Reviewed against master's parser source rather than taking the description on faith — the cascade extension is correct. Group mappings for LikeDeclRegex/EquateDeclRegex/ClassInstanceDeclRegex all check out, the refactor preserves the old ref/scalar semantics (including PRIVATE exclusion) exactly, and the catch-all is properly guarded: bare `Init PROCEDURE` prototypes are filtered by the keyword list, and parenthesized prototypes can't match `ClassInstanceDeclRegex` at all. The comment about GROUP/QUEUE/RECORD lines being intercepted before the cascade is accurate.

Independently verified by rebuilding the indexer and re-running the fixture: 18 callers of `WorkerClass.Sign`, `LikeMemberBugClass.GenCertData` (`LIKE(SmallGroupType)`) and `.SomeHandle` (`SMALLHANDLETYPE`) both captured with `scope='class'`, 7 class symbols, no regressions on #79–#90.

One small thing found and fixed (pushed to this branch as 51193b8, maintainer edit): the two lines inserted into `Worker.clw` shift `MainHelperProc`'s `worker.Sign(5)` call from line 56 to 58, but the README caller table still said 56.

Note for merging: this PR and #93 conflict in every fixture file (both bump 17→18 callers, both add a class at the same spot in `WorkerLib.inc`, both edit the same README lines). Whichever lands second needs a rebase reconciling the counts (19 callers, 8 classes, merged SQL name list) and a fixture re-run.

Issue #89, opened 2026-07-14
CodeGraph:  GROUP(NamedType) local DATA-section declarations never captured — silently missing,


geircodes:
## Summary

Investigating why a procedure with an unusually large local-variable list indexed incompletely turned up two local variables that were missing entirely — not misattributed, not misclassified, just absent, with zero trace anywhere in the database. Both were declared as a named, parenthesized `GROUP` instantiation (`Name GROUP(NamedType)`, each followed by its own `END` on a separate line) — ordinary, common Clarion syntax for a local variable whose fields are entirely inherited from a named `GROUP,TYPE`. Every other local variable in the same procedure, including ones declared immediately afterward, indexed correctly.

This is a different bug from a previously-found one involving the same construct in a different context (a `.inc` file's `CLASS` body): that one was a cascading state-tracking leak. This one is narrower — the declaration line simply matched none of the parser's declaration patterns and produced nothing at all, with no effect on anything declared after it.

---

## Root cause

`ParseMemberFile` (the parser for procedure implementations in `.clw` files) tries six different declaration patterns, in order, against each line inside a procedure's `DATA` section, to figure out what kind of variable a line declares. The one meant for `GROUP`/`QUEUE` declarations required the keyword to be followed immediately by either the end of the line or a comma-attribute list (e.g. `,PRE(xx)`). A parenthesized inherited-type argument right after the keyword — `Name GROUP(NamedType)` — didn't fit that shape.

None of the other five patterns covered it either: they all require the type token to be followed only by a comma-clause or the end of the line, never an opening parenthesis. So the line matched nothing, and the variable was never created as a symbol — silently, with no error and no trace, the same failure signature every gap in this codebase's parser has had so far.

Because nothing matched, none of the parser's internal state was touched either — no depth counter was incremented, nothing was left in an inconsistent state for later lines to trip over. That's what makes this bug narrower than a related one found earlier in a different file type: there, the equivalent construct's failure mode corrupted a state counter and took out everything declared afterward. Here, only the two specific declarations vanished; their neighbors were fine both before and after the fix.

---

## Fix

Widened the `GROUP`/`QUEUE` declaration pattern to accept an optional parenthesized type argument between the keyword and the optional comma-attribute suffix.

[...]

PR #93, opened 2026-07-15
fix(codegraph): capture GROUP-instantiated CLASS members' own names

geircodes:
# CLASS member that is itself a GROUP instantiation never gets its own name captured as a symbol

## Summary

A CLASS member declared as a `GROUP` instantiation — whether self-closing on one line (`InlineGroup GROUP(SomeType) END`) or spread across multiple lines with its own extra fields — never got a symbol for its own name at all. This is distinct from #88: that fix corrected a depth-tracking leak affecting *every other* member declared afterward in the class, but never added actual symbol creation for the triggering member itself. After #88, the member causing the leak still silently didn't exist as a symbol — only its side effects on later members were fixed.

## Root cause

The check that tracks nesting depth for this construct (`ParseIncFile`'s "Nested END for inner GROUP/QUEUE etc.") only ever matched the line's shape with a loose, non-capturing regex, purely to manage a depth counter. It never extracted the member's name or type, and never created a symbol. That existing check already matched `QUEUE`/`RECORD` shapes too (inherited from #88, for depth-tracking purposes), so the fix below naturally covers all three for consistency — though `GROUP` is the only one realistically used this way in practice: `QUEUE` cannot be a plain by-value CLASS member at all (must be `&QUEUE`, confirmed directly against a real compile), and `RECORD` is conventionally used for file/database record structures, not CLASS composition.

## Fix

Replaced the loose shape check with one that captures the member's name, keyword, and optional named-type — mirroring the `GROUP/QUEUE(NamedType)` local-variable fix (#89). Creates a `variable` symbol for direct class members, respecting the same `PRIVATE`-attribute exclusion the simple-member cascade already uses. Self-closing detection now uses the regex's own capture group instead of a separate re-check.

## Minimal reproduction

Added to `test-fixtures/codegraph-repro/WorkerLib.inc` — a class whose members cover both the genuine multi-line form (with its own extra field) and `PRIVATE`-attribute exclusion for this construct:

```clarion
MultiLineGroupBugClass CLASS,TYPE, MODULE('WorkerLib.CLW'), LINK('WorkerLib.CLW', 1 ), DLL( 0 )

HiddenGroupMember      GROUP(SmallGroupType), PRIVATE
                       END
MultiLineGroup         GROUP(SmallGroupType)
ExtraField                LONG
                       END
AfterMultiLineGroupMember &WorkerClass
CallViaAfterMultiLineGroupMember PROCEDURE( ), LONG
            END
```

The existing fixture already has two other variants of the same construct, previously added only to prove #88's depth-tracking fix — neither ever produced a symbol for its own name until this PR:

```clarion
GroupBugClass CLASS,TYPE, MODULE('WorkerLib.CLW'), LINK('WorkerLib.CLW', 1 ), DLL( 0 )

HiddenMember            &WorkerClass, PRIVATE
InlineGroup             GROUP(SmallGroupType) END
AfterGroupMember        &WorkerClass
CallViaAfterGroupMember PROCEDURE( ), LONG
            END

PeriodBugClass CLASS,TYPE, MODULE('WorkerLib.CLW'), LINK('WorkerLib.CLW', 1 ), DLL( 0 )

InlineGroupPeriod        GROUP(SmallGroupType) .
AfterPeriodMember        &WorkerClass
CallViaAfterPeriodMember PROCEDURE( ), LONG
            END
```

**Before this fix**: `GroupBugClass.InlineGroup`, `PeriodBugClass.InlineGroupPeriod`, and `MultiLineGroupBugClass.MultiLineGroup` do not exist as symbols anywhere, despite each opening and (in the last case) closing correctly and being fully compilable Clarion. `HiddenGroupMember` is also absent, but for the wrong reason (no symbol creation existed for this construct at all yet, so its `PRIVATE` exclusion was never actually exercised).

**After this fix**: all three members correctly exist as `variable` symbols (`params='GROUP(SmallGroupType)'`, `scope='class'`), `HiddenGroupMember` correctly stays absent (now genuinely excluded by its `PRIVATE` attribute), and every neighboring member (`AfterGroupMember`, `AfterPeriodMember`, `AfterMultiLineGroupMember`) remains correctly captured — proving no new depth-tracking regression.

## Verification

Extended `test-fixtures/codegraph-repro` with `MultiLineGroupBugClass` to cover the genuine multi-line `GROUP` form and `PRIVATE`-attribute exclusion for this construct. `GroupBugClass.InlineGroup` and `PeriodBugClass.InlineGroupPeriod` (both already in the fixture for #88) now retroactively get correct symbols with no fixture changes needed for those two. 18 total callers of `WorkerClass.Sign`, up from 17. No regressions on any prior fix.

Also verified against a real production solution: 23 previously-invisible class-scope `GROUP`-typed members found solution-wide — including one on the very same class that originally motivated #88 (its first property), which #88's depth-leak fix never actually made visible itself.

### DIFF
--- a/ClarionAssistant/CodeGraph/Parsing/ClarionParser.cs
+++ b/ClarionAssistant/CodeGraph/Parsing/ClarionParser.cs
@@ -79,6 +79,16 @@ namespace ClarionCodeGraph.Parsing
             @"^([\w:]+)\s+(GROUP|QUEUE)\s*(\([^)]*\))?\s*(?:(,.*)|(?<term>END\b\s*|\.\s*))?$",
             RegexOptions.Compiled | RegexOptions.IgnoreCase);
 
+        // CLASS member that is itself a GROUP/QUEUE/RECORD instantiation (also allows RECORD,
+        // unlike GroupQueueDeclRegex above, which only ever needed GROUP/QUEUE for DATA-section
+        // locals). Same shape and same named "term" self-closing group as GroupQueueDeclRegex --
+        // used by ParseIncFile's CLASS-body nested-structure check to ALSO capture a symbol for
+        // the member's own name, not just track nesting depth (issue: GROUP/QUEUE/RECORD CLASS
+        // member's own name never captured as a symbol at all).
+        private static readonly Regex ClassGroupQueueRecordDeclRegex = new Regex(
+            @"^([\w:]+)\s+(GROUP|QUEUE|RECORD)\s*(\([^)]*\))?\s*(?:(,.*)|(?<term>END\b\s*|\.\s*))?$",
+            RegexOptions.Compiled | RegexOptions.IgnoreCase);
+
         // LIKE declaration: VarName LIKE(OtherVar) [,attributes]
         private static readonly Regex LikeDeclRegex = new Regex(
             @"^([\w:]+)\s+LIKE\s*\(([^)]+)\)\s*(,.*)?$",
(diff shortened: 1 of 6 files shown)


ClarionLive (approved):
The symbol-capture logic is right: depth-1 gating correctly limits capture to direct members, nested inner GROUPs still get depth tracking without spurious symbols, and the `,\s*PRIVATE\b` check is actually stricter than the cascade's `IndexOf` (which would false-positive on `,PRE(PRIVATE)`). The gate widening from `^\w+` to `^([\w:]+)` also incidentally fixes a latent bug — a colon-prefixed GROUP member previously missed the depth branch entirely, letting its inner END prematurely close the class.

One regression found, confirmed empirically, and fixed (pushed to this branch as ae9805b, maintainer edit): replacing the old end-of-line self-closing check with the regex's `term` group broke the **attrs + same-line-terminator** forms — `X GROUP(Type),DIM(2) END` / `X GROUP(Type),DIM(2).` — because the `,.*` attrs alternative swallows the terminator, so `term` never matches and `classEndDepth` leaked (the exact #88 failure class, which master handles correctly today). The fix keeps the regex for the gate + symbol capture but restores the end-of-line terminator re-check for the depth decision, in both synced parser copies, and adds `AttrTermGroup`/`AttrTermGroupPeriod` to the fixture covering both forms.

Verified both ways by rebuilding the indexer against the extended fixture: pre-fix parser drops to 17 callers and loses `AttrTermGroupPeriod` + `AfterMultiLineGroupMember` entirely; with the fix, 18 callers, both new members captured (`GROUP(SmallGroupType)`, `scope='class'`), `HiddenGroupMember` correctly absent, no regressions on #79–#90.

Heads-up: the same `term`-group limitation exists in `GroupQueueDeclRegex` on the #89 DATA-section local path (`dataGroupDepth`) — preexisting, not introduced here, and the same one-line fix applies. Worth a small follow-up.

Note for merging: conflicts with #92 in every fixture file — whichever lands second needs the counts reconciled (19 callers, 8 classes, merged SQL name list) and a fixture re-run.

PR #94, opened 2026-07-15
fix(diff): detect file encoding instead of assuming UTF-8 in show_diff

geircodes:
# show_diff misdecodes special characters when loading original_file/modified_file from disk

## Summary

`show_diff`'s disk-loading paths (`original_file`/`modified_file` params) read files via `File.ReadAllLines()` with no encoding argument. That overload defaults to UTF-8 when no byte-order mark is present. Clarion source in this project (and many others) is traditionally saved as Windows-1252 with no BOM, so any special character (Norwegian `æøå`, `Ü`, `§`, em-dash, etc.) renders as a replacement character (`<?>`) in the diff viewer, and the mis-decode can desync the rest of that line's display (observed: a trailing `|` line-continuation character disappearing entirely). This makes the diff viewer produce false-positive "differences" on lines that are actually identical between disk and buffer.

## Root cause

`DiffService.ShowDiffFromFile` and the private `ReadFileRange` (used by `ShowDiffFromFiles`) both call `File.ReadAllLines(path)` — the 1-arg overload, which only detects a UTF-8/UTF-16 BOM and otherwise assumes UTF-8. A correct BOM-sniff-with-ANSI-fallback helper already existed (`ModernEmbeditorViewContent.DetectFileEncoding`), but it was private to that file and never used by `DiffService`.

A pure BOM-sniff-else-ANSI fallback isn't quite right either, though: this repo's own `.cs` files are frequently genuine UTF-8 *without* a BOM (e.g. `DiffService.cs` and `ModernEmbeditorViewContent.cs` itself — the latter has ~170 lines with em-dashes in comments). Falling back to ANSI unconditionally on any no-BOM file would fix Clarion source but mangle those.

## Fix

- Extracted `DetectFileEncoding` out of `ModernEmbeditorViewContent` into a new shared `Services/EncodingHelper.cs` (`internal`/`public static`), and updated `ModernEmbeditorViewContent`'s two call sites to use it (no behavior change there — same logic, just de-duplicated).
- Hardened the no-BOM case: instead of assuming ANSI outright, attempt a strict UTF-8 decode of the file's bytes first (`new UTF8Encoding(false, throwOnInvalidBytes: true).GetString(...)`). A lone Windows-1252 high-bit byte (e.g. `Ü` = `0xDC`) is not a valid standalone UTF-8 sequence and throws, correctly falling back to ANSI. Genuine no-BOM UTF-8 content (e.g. an em-dash, `0xE2 0x80 0x94`) decodes cleanly and is kept as UTF-8. This is a single linear scan over bytes already read into memory — no measurable cost even on large files, since disk I/O dominates either way.
- `DiffService.ShowDiffFromFile`/`ReadFileRange` now pass `EncodingHelper.DetectFileEncoding(path)` explicitly to `File.ReadAllLines` instead of relying on the no-arg overload's UTF-8 default.

## Verification

Live-tested against a running Clarion 11.1 IDE session with the built addin deployed:

1. **Windows-1252 Clarion source** (`.clw` file with unsaved editor changes): before the fix, lines containing `Ü`/`ü`/`ø`/`§` showed as `<?>` on the disk side with a trailing `|` continuation character missing, falsely flagging those lines as different. After the fix, both sides render identically; only the developer's actual unsaved edit shows as a diff.
2. **UTF-8-without-BOM `.cs` source** (`ModernEmbeditorViewContent.cs` itself, ~170 lines containing em-dashes, with an unsaved one-character edit): confirmed the em-dash lines render identically on both sides (no mojibake), and only the actual edited line shows as a diff — proving the content-based UTF-8 detection correctly avoids the ANSI-fallback regression that a pure BOM-sniff approach would have introduced for this exact file.

No existing behavior changes for BOM-present files (UTF-8 BOM, UTF-16 LE/BE) — those paths are untouched.

## Follow-up commit: trailing newline preserved on whole-file diffs

`ShowDiffFromFile`/`ShowDiffFromFiles` reconstructed the disk side by splitting the file into lines (`File.ReadAllLines`) and rejoining with `"\n"`, which always drops any trailing newline the file actually had. A live editor buffer's raw text naturally keeps that trailing EOL, so diffing a whole file against its own live buffer showed a spurious "extra blank line" at the end that was never a real edit — found while live-testing `show_diff`'s new `modified_from_active_editor` mode (a separate PR) against real files.

Added a shared `ReadOriginalText` helper: when no line-range subset is requested (the common whole-file case), read the file's exact text via `File.ReadAllText` instead of reconstructing it line-by-line, preserving the trailing newline exactly. A genuine partial line-range still goes through line splitting, since some reconstruction is unavoidable there. Replaces the old `ReadFileRange` (used only by `ShowDiffFromFiles`) with the same shared helper, fixing the identical trailing-newline artifact on that path too.

Verified live: a whole-file diff against a live buffer no longer shows a false extra trailing-line change alongside the developer's real edit.

### DIFF
--- a/ClarionAssistant/ClarionAssistant.csproj
+++ b/ClarionAssistant/ClarionAssistant.csproj
@@ -180,6 +180,7 @@
     <Compile Include="Terminal\ModernEmbeditorViewContent.cs" />
     <Compile Include="Terminal\MonacoEditorControl.cs" />
     <Compile Include="Services\DiffService.cs" />
+    <Compile Include="Services\EncodingHelper.cs" />
     <Compile Include="Services\DocGraphService.cs" />
     <Compile Include="Services\InstanceCoordinationService.cs" />
     <Compile Include="Services\ClarionTraceService.cs" />
--- a/ClarionAssistant/Services/DiffService.cs
+++ b/ClarionAssistant/Services/DiffService.cs
@@ -89,7 +89,7 @@ namespace ClarionAssistant.Services
                 if (!File.Exists(originalFile))
                     return "Error: File not found: " + originalFile;
 
-                string[] allLines = File.ReadAllLines(originalFile);
+                string[] allLines = File.ReadAllLines(originalFile, EncodingHelper.DetectFileEncoding(originalFile));
 
                 // Clamp line range (1-based input)
                 if (startLine < 1) startLine = 1;
@@ -138,7 +138,7 @@ namespace ClarionAssistant.Services
 
         private static string ReadFileRange(string filePath, int startLine, int endLine)
         {
-            string[] allLines = File.ReadAllLines(filePath);
+            string[] allLines = File.ReadAllLines(filePath, EncodingHelper.DetectFileEncoding(filePath));
             if (startLine < 1) startLine = 1;
             if (endLine < 1 || endLine > allLines.Length) endLine = allLines.Length;
             if (startLine > endLine) startLine = 1;
--- a/ClarionAssistant/Services/DiffService.cs
+++ b/ClarionAssistant/Services/DiffService.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Text;
 using System.Windows.Forms;
 using ClarionAssistant.Terminal;
 using ICSharpCode.SharpDevelop.Gui;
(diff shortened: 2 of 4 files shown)


ClarionLive (approved):
Verified against master's source and compile-checked (builds clean as 5.3.754):

- The `endLine == -1` whole-file sentinel is correct for every existing caller (McpToolRegistry's show_diff defaults, MonacoFileOpener.Compare's explicit `1, -1`).
- The CRLF question the ReadAllText change raises checks out fine: the classic renderer's `SplitLines` splits on `\r\n|\n|\r` and Monaco handles mixed EOLs natively, so preserving the file's real EOLs produces no spurious per-line diffs in either renderer.
- Detection logic is sound (BOM sniff, strict-UTF-8 probe with ANSI fallback; `Encoding.Default` is the ANSI codepage on .NET Framework; StreamReader strips matched BOMs).
- The `start > end` change from a silent `startLine = 1` to a thrown ArgumentException is an improvement, and ShowDiffFromFiles' existing catch surfaces it as a clean error string.

One trade-off to be aware of (not blocking): keeping the disk side's trailing newline fixes the spurious blank-line diff against a live buffer, but an agent passing whole-file `modified_text` without a trailing `\n` will now see a trailing empty-line change that used to be invisible. Technically a real difference — hiding it was arguably the old bug — but expect that cosmetic row in agent-driven whole-file diffs.

PR #95, opened 2026-07-15
feat(terminal): support pasting a clipboard image as an @-reference file

geircodes:
# Pasting a clipboard image in the terminal does nothing — no way to hand Claude a screenshot

## Summary

`OnPasteRequested` only checked `Clipboard.ContainsText()`. Screenshot tools (Snipping Tool, Win+Shift+S) put a bitmap on the clipboard with no accompanying text, so pasting a screenshot into the docked terminal silently did nothing — no error, just no visible effect. The only workaround was: save the screenshot to disk manually, find its path, and type an `@path` reference by hand. Tedious, and the saved files just piled up with nothing to clean them up afterward.

This matters beyond convenience: some content simply can't be copied as text at all (e.g. a hex-viewer's rendered byte grid, a rendered UI glitch, a diagram) — a screenshot is the *only* way to hand that to an assistant, so this wasn't a nice-to-have, it was a dead end for an entire class of "show, don't describe" situations.

## Fix

Mirrors the existing large-text-paste-to-file path (`WritePasteCacheFile`): when the clipboard has no text but `Clipboard.ContainsImage()` is true, save the bitmap as PNG to the same `paste-cache` directory and inject an `@<path> ` reference the same way. Claude Code reads image files natively via `@reference`, so no additional plumbing was needed on that side.

Refactored the cache-directory-resolution and 24h-prune-sweep logic (previously inlined in `WritePasteCacheFile`) into two small shared static helpers (`PasteCacheDir`, `PrunePasteCache`) so the new `WriteImagePasteCacheFile` doesn't duplicate that setup — the prune sweep now covers both `paste-*.txt` and `paste-*.png`, so screenshots get the same automatic cleanup the text cache already had.

Also enabled the right-click context menu's "Paste" item when the clipboard holds an image, not just text (it previously only checked `ContainsText()` for its `Enabled` state).

## Verification

Live-tested in the docked terminal: pasted a screenshot (a Clarion source excerpt from an IDE window), confirmed it was written to `paste-cache` as PNG, the `@path` reference was injected into the prompt automatically, and Claude correctly read and described the image's content on submit.

### DIFF
--- a/ClarionAssistant/Terminal/WebViewTerminalRenderer.cs
+++ b/ClarionAssistant/Terminal/WebViewTerminalRenderer.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Drawing;
+using System.Drawing.Imaging;
 using System.IO;
 using System.Reflection;
 using System.Text;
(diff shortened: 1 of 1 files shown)


ClarionLive (approved):
Clean mirror of the existing large-text-paste path — compile-checked (builds clean as 5.3.754). `using (image)` disposes properly, `GetImage()` null is handled, the context-menu enablement update is right, and the text-path restructure preserves the old behavior exactly, including the cache-write-failure fall-through.

Two minor follow-up nits, neither blocking:
- Prune is per-pattern per-writer: text pastes prune only `paste-*.txt`, image pastes only `paste-*.png`, so a lone screenshot lingers past 24h until the next image paste. Pruning both patterns in both writers would be a two-line change.
- `Clipboard.GetImage()` can throw on clipboard contention and isn't wrapped — same preexisting exposure `GetText()` already has, so not a regression, but worth folding into any future hardening pass.

PR #96, opened 2026-07-15
feat(diff): optional current-line comparison panel in the Monaco diff viewer

geircodes:
# Add an optional current-line comparison panel to the Monaco diff viewer

## Summary

In the side-by-side/inline diff view, a long line gets cut off by the pane width, and it can be hard to spot exactly which characters differ within a changed line from the inline highlight alone. Inspired by TotalCommander's "Compare Files" feature, which shows the current line from both sides in a dedicated full-width field at the bottom with the differing part highlighted.

## Fix

Added a "Line View" toolbar toggle (off by default, persisted via `localStorage` like the other toolbar toggles — Inline/Side-by-side, Ignore WS, theme, font). When enabled, a panel below the editor shows the full text of the currently navigated change's original ("Disk") and modified ("Buffer") lines, with the exact differing character range highlighted on each side.

No new diff computation was needed: Monaco's diff editor already computes and exposes per-change character-level diffs via `getLineChanges()[].charChanges` (start/end column for the differing substring on each side), which this reuses directly.

**Scope note:** deliberately tied to the existing Prev/Next change navigation, not free cursor tracking. A navigated change already pairs up the corresponding original/modified line numbers and their `charChanges`, so no separate line-number-mapping logic is needed. Matching TotalCommander exactly (panel follows the cursor to *any* line, including unchanged ones) would need extra logic to map an arbitrary line's position across inserted/deleted regions elsewhere in the file — correctly computing that requires walking the full change list and accumulating line-count offsets. Deferred as a possible follow-up; Prev/Next-tied covers the primary use case (inspecting a specific difference) without that complexity.

Also auto-navigates to (reveals + focuses) the first change when the diff opens, instead of silently starting on "change 0" while the editor viewport stays on line 1 — the Line View panel would otherwise show content for a change the visible editor hadn't scrolled to yet, and pressing Next would skip straight to the *second* change. Now the viewport, counter, and Line View panel are all consistent with "1 of N" from the moment the diff loads, matching what Next/Prev already do.

## Verification

Live-tested in a running Clarion 11.1 IDE: toggling "Line View" shows both lines with the changed character range highlighted correctly, including with multiple separate changes on the same line; navigating via Prev/Next updates the panel to match; the panel is already populated on open without needing to press Next first; the editor area resizes properly when the panel opens/closes (Monaco's `automaticLayout` picks up the container resize on its own via the `--lc-height` CSS variable toggle).

### DIFF
--- a/ClarionAssistant/Terminal/monaco-diff.html
+++ b/ClarionAssistant/Terminal/monaco-diff.html
@@ -62,7 +62,10 @@
     #btn-theme { min-width: 28px; padding: 3px 8px; font-size: 14px; line-height: 1; }
 
     /* ---------- Editor area ---------- */
-    #diff-editor { width: 100%; height: calc(100% - 38px - 44px); }
+    /* --lc-height is toggled by JS (0px / panel's real height) so the editor always fills the
+       remaining space exactly — see toggleLineCompare(). automaticLayout:true on the Monaco diff
+       editor picks up the resulting container resize on its own (ResizeObserver). */
+    #diff-editor { width: 100%; height: calc(100% - 38px - 44px - var(--lc-height, 0px)); }
     #loading { position: absolute; top: 50%; left: 50%; transform: translate(-50%,-50%); color: var(--counter-color); font-size: 14px; }
 
     /* ---------- Bottom action bar ---------- */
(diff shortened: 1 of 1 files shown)


ClarionLive (approved):
Verified the fiddly parts: `highlightLine`'s column math is correct for Monaco's 1-based end-exclusive charChanges convention (sorted, overlap-safe), and all line content is HTML-escaped before hitting innerHTML — no injection path from file content. Missing charChanges degrades gracefully to unhighlighted lines. The `--lc-height` CSS-variable + automaticLayout resize approach is sound, persistence matches the other toolbar toggles, and the auto-reveal-first-change fix is properly guarded (`currentChangeIndex < 0`) so it fires only on the first diff computation. Script syntax-checked clean.

One cosmetic follow-up nit: `refreshChanges` doesn't clamp `currentChangeIndex`, so if toggling Ignore-WS shrinks the change list, `updateLineCompare` gets undefined and early-returns, leaving the previous lines visible in the panel while the counter says otherwise. Clearing the two spans when `change` is null would cover it.

PR #99, opened 2026-07-15
feat(embeditor): dictionary- and CodeGraph-aware completion in the CA Embeditor

### LINKED ISSUE #98: Wire SchemaGraph (dictionary tables/fields/keys) into CA Embeditor completion
SchemaGraph already indexes dictionary tables/columns/keys into `.schemagraph.db` (Schema Sources panel, `query_schema`/`search_tables`/`get_table`/`search_columns` MCP tools), but none of that reaches the CA Embeditor's Monaco completion — only LSP + CodeGraph symbols do. Typing a table's `PRE:` prefix (e.g. `Cus:`) never suggests dictionary fields or keys, and bare-prefix completion never suggests table names.

Separately, completion `Detail` text for variables is misleading/incomplete today:
- CodeGraph-indexed global variables show the literal word "variable" instead of the declared type — `CgCompletionDetail` never reads `Params` (which holds the actual type).
- No variable completion source distinguishes local vs. global scope consistently, or surfaces a `!`-comment as a description.

A PR follows.

asantarelli:
Closes #98

## Summary
- Wires SchemaGraph into the CA Embeditor's completion pipeline: dictionary **fields and keys** via a table's `PRE:` qualifier, and bare-prefix **table name** completion — both additive, shown alongside (not instead of) same-named in-buffer `GROUP`/`QUEUE` fields.
- Fixes CodeGraph-global variable completion showing the literal word "variable" instead of the declared type (`Params` was never read).
- Normalizes variable scope wording to local/global across all four variable-completion sources, and surfaces a trailing `!` comment as description.
- Dictionary field completion now shows the declared `TYPE(size[,places])`, not the picture (which is a display mask, not a size — e.g. a `DECIMAL(13,2)` with picture `@n$4.2` was showing `@n$4.2` as if it were the size).
- Long completion text (descriptions, key composition) moved from `Detail` (single line, truncates) to `Documentation` (Monaco's expandable, wrapping docs panel).

## Test plan
- [ ] Add + index a Schema Source (dictionary) via Solution Settings → Schema Sources
- [ ] In the CA Embeditor, type `<TablePrefix>:` → dictionary fields and keys appear, with type/description/composition in the docs panel
- [ ] Confirm a same-named local `GROUP,PRE(...)` still shows its own fields alongside the dictionary ones
- [ ] Type a table-name prefix bare (no `:`) → table name suggested
- [ ] Confirm local/module/global variable completion shows declared type + scope, and a `!` comment shows as description in the docs panel

### DIFF
--- a/ClarionAssistant/Services/McpToolRegistry.cs
+++ b/ClarionAssistant/Services/McpToolRegistry.cs
@@ -69,6 +69,9 @@ namespace ClarionAssistant.Services
             // SharedLspBridge answer cross-project definition/references/workspace-symbol in C#
             // when the LSP returns nothing (needed once we adopt Mark's pure upstream server.js).
             SharedLspBridge.CodeGraphDbPathProvider = () => FindCodeGraphDb();
+            // Same wiring for the dictionary schema DB — lets SharedLspBridge's completion merges resolve
+            // the same .schemagraph.db that query_schema/search_tables/get_table already use.
+            SharedLspBridge.SchemaGraphDbPathProvider = () => FindSchemaGraphDb();
         }
 
         public void SetDiffService(DiffService diffService)
--- a/ClarionAssistant/Services/SchemaGraphService.cs
+++ b/ClarionAssistant/Services/SchemaGraphService.cs
@@ -2067,6 +2067,211 @@ namespace ClarionAssistant.Services
             }
         }
 
+        /// <summary>
+        /// Dictionary-aware qualifier completion for the CA Embeditor: given a table PREFIX (e.g. "Cus"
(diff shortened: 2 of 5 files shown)


ClarionLive (approved):
Reviewed line-by-line against the schema DDL, the completion merge pipeline, and the Monaco serialization path — all checks out. Verified locally with a clean MSBuild Debug build. Pushed one small addition onto the branch: `lspKindToMonaco` had no `case 20`, so the dictionary KEY items (Kind=20, EnumMember) were falling through to the generic Text icon — now mapped. Nice work, thanks for the thorough comments throughout.

PR #104, opened 2026-07-16
feat(embeditor): in-editor find & replace — top-right overlay + Find-All column (#66)

### LINKED ISSUE #66: Change search bottom-bar to a floating window.
When performing a Search, it opens a "bottom-bar" (panel on the lower part of the editor window) that occupies too much of the screen. This should be a floating window, so you can move it out of the way.

msarson:
## Summary

Replaces the wide, bottom-docked Find & Replace panel with an **in-editor find/replace modelled on VS Code**, addressing #66. Feedback on the webinar (and from the issue's reporter) was that the docked "search bottom-bar" ate editor space and the eager "find-all" grid was noise. This reworks it into a compact quick-find overlay plus an opt-in Find-All column, while **keeping the persisted, cross-tab, per-procedure search history** the existing panel already had.

Two files change: `MonacoEditorControl.cs` (a focused host fix) and `monaco-embeditor.html` (the UI).

## Why not just use Monaco's native find widget?

I prototyped exactly that first (Ctrl+F → `editor.getAction('actions.find')`). It looks and feels right, but it can't be backed by our history store: Monaco's find widget keeps its own in-memory, per-instance history, and because every embeditor tab is a **separate WebView2 with its own Monaco runtime**, that history is per-tab and dies on tab close. VS Code's shared/persisted find history comes from workbench services Monaco doesn't ship. Since we already persist and broadcast search history host-side (`saveHistory`), the native widget would have been a regression on the one thing we care about. So I kept our engine + history and rebuilt the *presentation* to match VS Code. The native-find experiment is reverted; the two bug fixes it surfaced are kept.

### 1. Quick find — Ctrl+F / Ctrl+H
- Compact overlay pinned to the editor's **top-right**, floating over the code — **reserves no editor space** (the core #66 complaint).
- **Incremental** search-as-you-type, landing on the **first match at/after the caret** (not always match #1).
- **Enter / Shift+Enter** and inline **‹ ›** buttons cycle matches.
- **↑ / ↓** walk the **persisted** history — works on reopen, and is shared across tabs.
- VS Code-style **boxed input with the toggles inside it** (`Aa` case, `ab` word, `.*` regex, `🔒` include-read-only), borderless with an active highlight.
- Match count drops onto its **own line under the box**, so the control row stays compact; deterministic sizing so nothing clips.

### 2. Find All — Ctrl+Shift+F / Ctrl+Shift+H
- Opt-in **left-docked column** in a vertical VS Code Search-view layout: slim header (title · count · close) → full-width search/replace → results list.
- **Ellipsized, fixed-layout results** that can't widen the panel; **click a row to jump**. Currently scoped to the open file (Monaco sees one buffer) — this is the natural surface for a future solution-wide search.

### 3. Replace — VS Code expander
- A **left chevron** expands/collapses the replace row (rotates when open).
- Replace **only auto-opens on the replace chords** (Ctrl+H / Ctrl+Shift+H); the find chords open collapsed.

### 4. File-mode tidy-up
- The "include read-only (generated) blocks" toggle is **hidden in file mode**, where the whole buffer is editable and it has no meaning.

### A. Find accelerators leaked to the hidden native editor — `MonacoEditorControl.cs`
The Modern embeditor docks its WebView2 **on top of Clarion's still-open native embeditor** (chrome hidden, editor alive as the write-back target). With `AreBrowserAcceleratorKeysEnabled = false`, the WebView delivers find keys to the DOM **and also forwards them to the host**, where SharpDevelop's form-level `ProcessCmdKey` ran its Find / Find-in-Files / Replace against the active view content — the hidden native editor — popping its search bar up under ours. `MonacoEditorControl` sits below the workbench form in the parent chain, so overriding `ProcessCmdKey` and returning `true` for the find chords **while the WebView has focus** consumes the forwarded key first. Focus-scoped, so normal Clarion editors are unaffected. (Verified via the decompiled IDE that this is form-`ProcessCmdKey` menu-shortcut dispatch, not the global `IMessageFilter`, which only handles Escape.)

### B. Replace All was a no-op in slot mode — `monaco-embeditor.html`
The read-only guard reverted any edit event that touched a generated (read-only) line **wholesale**. Replace All is one atomic event whose matches land in both editable slots and generated code, so the legal in-slot replacements were reverted too. The guard now reverts the batch, then **re-applies only the changes that fell inside editable slots** (also salvages multi-cursor edits straddling a boundary).

## Implementation notes
- Built on the existing `runFind` / `model.findMatches` / decorations and the `saveHistory` host bridge — **no new host plumbing** for the find UI.
- The overlay and the column are two CSS modes of one panel; all the search/replace/history machinery is shared.
- History is committed to the persisted store on Enter / close (not per keystroke), so incremental typing doesn't pollute it.

## Testing (manual, in the Clarion 11.1 IDE)
- Ctrl+F / Ctrl+H / Ctrl+Shift+F / Ctrl+Shift+H in an open procedure embed — no native search bar leaks underneath.
- Incremental find-from-cursor, next/prev, and ↑/↓ history recall including after closing and reopening the embeditor.
- Replace All in slot mode with matches in both slots and generated code — only the editable ones change; read-only skipped and reported.
- Find-All column at narrow dock widths — no overflow into the editor.
- File mode vs slot mode (🔒 toggle presence) and light/dark themes.

Closes #66

### DIFF
--- a/ClarionAssistant/Terminal/MonacoEditorControl.cs
+++ b/ClarionAssistant/Terminal/MonacoEditorControl.cs
@@ -373,6 +373,35 @@ namespace ClarionAssistant.Terminal
             catch { }
         }
 
+        // ── Keep editor accelerators inside Monaco (issue #66 / native-find leak) ────────────────
+        // OVERLAY REALITY: this Monaco WebView2 is docked FILL on top of Clarion's still-open native
+        // embeditor (ClaGenEditor) — its chrome is only hidden (ModernEmbeditorViewContent.HideNativeChrome),
+        // the native text editor is alive underneath as the write-back target. With
+        // AreBrowserAcceleratorKeysEnabled=false the WebView still delivers Ctrl+F to the DOM (our page
+        // opens Monaco's own find), but it ALSO forwards the accelerator up to the host. SharpDevelop's
+        // form-level ProcessCmdKey (DefaultWorkbench) then runs its Find menu-shortcut against the ACTIVE
+        // view content — the hidden native editor — so its search bottom-bar leaks up on top of Monaco's.
+        // This control sits BELOW the workbench form in the parent chain, so returning true here consumes
+        // the forwarded key before the form can dispatch. The DOM keydown already reached the page, so we
+        // only swallow the host-side leak; we don't re-trigger the find here.
+        protected override bool ProcessCmdKey(ref Message msg, Keys keyData)
+        {
+            if (_webView != null && _webView.ContainsFocus)
+            {
+                switch (keyData)
+                {
+                    case Keys.Control | Keys.F:                 // Find
+                    case Keys.Control | Keys.Shift | Keys.F:    // Find All (Ctrl+Shift+F) — else it leaks to Find-in-Files
+                    case Keys.Control | Keys.H:                 // Replace
+                    case Keys.Control | Keys.Shift | Keys.H:    // Find All + Replace (Ctrl+Shift+H) — else leaks to Replace-in-Files
+                    case Keys.F3:                               // Find next
+                    case Keys.Shift | Keys.F3:                  // Find previous
+                        return true;               // handled — don't let the workbench run its native Find
+                }
+            }
+            return base.ProcessCmdKey(ref msg, keyData);
+        }
+
         /// <summary>Reply to an LSP/request message: {type:"response", reqId, data}.</summary>
         public void PostResponse(int reqId, IDictionary<string, object> data)
         {
--- a/ClarionAssistant/Terminal/monaco-embeditor.html
+++ b/ClarionAssistant/Terminal/monaco-embeditor.html
@@ -304,30 +304,116 @@
     :root { --find-accent: #89b4fa; --find-hit: rgba(137,180,250,0.35); --find-hit-border: #89b4fa; --on-fg: #1e1e2e; --find-fz: 12px; }
     body.light { --find-accent: #1e66f5; --find-hit: rgba(30,102,245,0.22); --find-hit-border: #1e66f5; --on-fg: #eff1f5; }
 
+    /* QUICK-FIND mode (default when open): compact overlay pinned to the editor's top-right corner,
+       reserves NO editor space — it floats over the code (VS Code-style). `top` is set in JS
+       (positionFindPanel) to sit just under the toolbar; the rest is CSS. */
     #findPanel {
-        position: absolute; left: 0; right: 0; bottom: var(--status-bar-h); z-index: 15;
-        height: 300px; display: flex; flex-direction: column;
-        background: var(--toolbar-bg); border-top: 2px solid var(--find-accent);
+        position: absolute; right: 20px; left: auto; bottom: auto; z-index: 50;
+        width: auto; max-width: 94%; display: flex; flex-direction: column;
+        background: var(--toolbar-bg); border: 1px solid var(--find-accent); border-radius: 7px;
+        box-shadow: 0 4px 16px rgba(0,0,0,0.32);
         font-size: var(--find-fz); color: var(--fg);
     }
     #findPanel.hidden { display: none; }
-    #findGrip { height: 5px; cursor: ns-resize; background: var(--toolbar-border); flex-shrink: 0; }
+    /* Quick mode hides the results-grid machinery, the row labels and the standalone Find button. */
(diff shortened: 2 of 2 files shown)


msarson:
For traceability: the two embeditor-overlay bug fixes bundled in this PR are now tracked in #105 — (1) find accelerators leaking to the hidden native editor, and (2) Replace All no-op'ing in slot mode. Keeping them here rather than splitting into a separate PR, per discussion.

(If you'd rather it auto-close on merge, I can promote this to a `Closes #105` line in the description — left the PR body as-is for now.)

ClarionLive:
Heads-up before you sink more time into this, Mark — there's a collision we're sorting out from this end.

The CA Find/Replace **dockable pad** landed on master yesterday (`c0a8665` + fixes, GitHub #66), and it's deployed at 5.3.773. It answers the same complaint from the other direction: find/replace moved out of the editor page entirely into a SharpDevelop pad (`CaFindPad` + `Terminal/ca-find.html`), with `monaco-embeditor.html` keeping the match/replace/decoration engine and the pad driving it remotely over a `caFind*` command channel. As part of that, the bottom `#findPanel` markup, CSS and its panel-era JS (`openFindPanel`, `renderResults`, `persistHistory`, the tab machinery) were **removed** from the page.

#104 branched from `52dd6dc`, before that landed, so it doesn't see any of it. GitHub still shows this PR as `MERGEABLE`, but that status is stale — a test merge against current master conflicts across 8 hunks in `monaco-embeditor.html`, because your work restyles a panel master no longer has. No need to rebase — you'd hit that wall too.

**Where John wants to land:** ship *both* and let the developer pick, rather than one replacing the other. The good news is they're far more compatible than the conflict suggests — both front-ends drive the *same* engine (`runFind` / `doReplace` / `gotoMatch` / `revealMatch` / `paintHighlights`). So it's one engine with two presentations: your floating overlay + Find-All column, or the dockable pad, chosen by a setting (`ClarionAssistant.FindUiMode`, defaulting to the pad since that's what's shipped).

**I'm doing the integration now**, since the pad side is fresh in my head — porting your presentation layer from `f5f5b8f` onto master rather than retyping it, so your authorship survives in the history. Two things from #104 I'm taking regardless of mode:
- the `ProcessCmdKey` accelerator swallow (`bd7073c`) — master has no equivalent, and the pad path needs it just as much to stop Ctrl+F reaching the native editor underneath;
- your incremental search-as-you-type and ↑/↓ history recall, which the pad doesn't do yet.

On history: your overlay persists via the `saveHistory`/`applyHistory` host channel (still intact on master); the pad has its own store in `ca-find-pad-state.json`. I'm going to unify them so both UIs share one history list — switching UI shouldn't lose your search history. Shout if you see a problem with that, or with any of the above.

ClarionLive:
Merged to master — via cherry-pick rather than the merge button, so GitHub won't mark this PR "merged", but both commits landed with your authorship intact: `227420b` (the `ProcessCmdKey` accelerator swallow) and `f134a5a` (the overlay + Find-All presentation, resolved against the pad work per the plan above).

What changed in the integration, so the diff against your branch makes sense:

- **Both UIs shipped, switchable**: Tools → Options → Clarion Assistant → Find / Replace picks pad vs overlay (`ClarionAssistant.FindUiMode`, default Pad since that's what was already deployed). Your overlay is exactly as you built it — one shared engine, two presentations.
- **Unified history**: the overlay's `saveHistory`/`applyHistory` channel and the pad's store now converge through one broker bus (`CaFindBroker.BroadcastHistory`) into the per-solution `find-history.json` — searching in either surface feeds the same `↑`/`↓` recall in both.
- **Review hardening**: the pipeline's cross-model pass found that with both surfaces visible they could desync (one showing query A while the engine held query B's matches — Replace All would rewrite the wrong set). Fixed with symmetric guards: replace always re-runs the surface's own visible query+options first (`9e2a617`, `98d762f`). Also `firstIdxFromCursor` is gated to overlay mode so pad-driven searches keep the shipped reveal-#0 behavior, and `closeFindPanel` cancels the pending incremental debounce.
- Your two bundled bug fixes are in (and #105 is closed): the accelerator swallow ships unconditionally — the pad path needed it just as much.

Released in 5.3.774. Closing since the content is on master; #66 is closed with the combined story. Thanks Mark — the overlay's a genuinely nice piece of work.

PR #110, opened 2026-07-16
feat(embeditor): document structure outline fly-out (Phase 1)

msarson:
## ⚠️ Stacked on #104

This branch is based on `feat/embeditor-in-editor-find` (#104) — it reuses that PR's left-column fly-out mechanism (`setEditorInsets` / the docked column). Until #104 merges, **this PR's diff includes #104's commits**; once #104 lands in `master` the diff reduces to just the outline work below. **Review #104 first.** The outline-specific commits are `f6f0c7f`…`e24e5a0` (files: `DocumentOutlineBuilder.cs`, `MonacoEditorControl.cs`, `ModernEmbeditorViewContent.cs`, `MonacoClarionSourceEditor.cs`, `SharedLspBridge.cs`, `ClarionAssistant.csproj`, and the outline section of `monaco-embeditor.html`).

## Summary

A **Document Structure outline** for the Monaco embeditor — a VS Code-style Outline shown as a left-column fly-out, driven by the LSP's `textDocument/documentSymbol`. First version (pull-based).

Toggle it from the new toolbar button (bulleted-list icon). It works in both host surfaces (the Modern embeditor and the source overlay); it's most useful in **file/source mode** (whole module, reliable symbols).

## What it does

- **Fly-out**: left column reusing #104's inset mechanism; opening it excludes the Find-All column and vice-versa.
- **Tree**: collapsible nodes (codicon chevron twisties); collapse state is keyed by node path and survives the debounced refreshes.
- **Class ▸ Methods regrouping**: a port of the extension's `StructureViewProvider.regroupFlatSymbols` — dotted method implementations (`Class.Method`, or `Class.Interface.Method`) are lifted from the flat root into a faux `Class ▸ Methods` (or `Class ▸ Interface`) tree, merged into the class' declaration node so declarations and implementations sit together, with each method's locals as children.
- **Icons**: VS Code **codicons** — Monaco already loads that font, so `codicon-symbol-<kind>` renders the exact Outline icons, with VS Code's symbol colours (method=purple, field/variable=blue, class/struct/enum=orange).
- **Toolbar** (replaces the title): **expand all**, **collapse all**, and a **live filter** (type e.g. `ThisWindow` to narrow to matching symbols + their ancestor path, force-expanded).
- **Horizontal scroll** instead of clipping — long names/signatures are fully readable.
- **Refresh**: on open and on a debounced buffer change (superseded-request-guarded). Pull-based; a server-push refresh is a Phase-2 follow-up.

## How the data gets here — LSP addin changes (ClarionLsp / clarion-lsp)

The outline needs the **hierarchical** `documentSymbol` tree. Two problems were fixed to deliver it through the C# path:

1. **`clarion-lsp` addin — flattening bug (released as v1.4.1).** `ClarionLspService.ParseDocumentSymbols` only read the top-level `result` array and dropped `children`, so a hierarchical `DocumentSymbol` (procedures with nested `WINDOW`/controls, classes with methods, routines with data) collapsed to just the root callables. It now **recurses `children`** and captures **`detail`** (control `USE(...)`, method signatures), reading the range from either `DocumentSymbol.range` or `SymbolInformation.location.range`. `SymbolResult` gained **`Children`** and **`Detail`**.
   - **First-load-safety**: `ClarionLsp.Contracts` stays `AssemblyVersion 1.0.0.0` and consumers ship a vendored copy, so the new fields are **written via reflection** in the addin (and read via reflection here). If an older contract wins the first-load race the addin degrades to a flat result rather than faulting — so **v1.4.1 is safe to install ahead of this CA change**.
   - Released as **[clarion-lsp v1.4.1](https://github.com/msarson/clarion-lsp/releases/tag/v1.4.1)**; the addin registry entry was bumped `1.4.0 → 1.4.1`.

2. **CA bridge — `SharedLspBridge.SymbolsToList`.** Now propagates the new `Children`/`Detail` into the dict tree the outline consumes, **read via reflection** so CA still compiles/loads against the older vendored contract — flat when absent, a full tree once the newer addin + contract are deployed. (The *bundled* LSP path already preserved the hierarchy; this fixes the *shared-addin* path, the common config.)

## Implementation (CA side)

- New **`documentStructure`** request action, modelled on the existing completion/definition handlers: `IMonacoEditorHost.OnDocumentStructure` → `MonacoEditorControl` dispatch → handlers in both hosts → `PostResponse({symbols, fileMode})`.
- **`Services/DocumentOutlineBuilder`** turns `DocumentSymbol[]`/`SymbolInformation[]` into `{name, kind, detail, line, children}`, mapping LSP 0-based lines → Monaco 1-based (embed-aware in slot mode, `+1` for whole-file source).
- All new host code is guarded and no-ops when the LSP isn't running.

## Scope / follow-ups

- **File/source mode** is the sweet spot. **Embed/slot mode** shows a partial outline by nature (the buffer is a procedure slice) — a dedicated slot navigator is a later phase.
- **Follow-cursor** (auto-select the symbol at the caret) and a **server-push refresh** (`clarion/symbolsRefreshed` already exists server-side) are natural next steps.

## Testing (manual, Clarion 11.1)

- Outline on a source file with classes/windows/routines: tree renders, method implementations grouped under `Class ▸ Methods` with their locals, icons + colours correct.
- Verified on the **shared addin path** (`Lsp.ForceLocal=false`) with the v1.4.1 addin — full hierarchy (previously flat); and confirmed the flat→tree difference isolates to the addin's `children` handling.
- Collapse/expand-all, filter (`ThisWindow`, a method name), click-to-navigate, horizontal scroll for long names, and refresh-on-edit while open.

### DIFF
--- a/ClarionAssistant/ClarionAssistant.csproj
+++ b/ClarionAssistant/ClarionAssistant.csproj
@@ -92,6 +92,7 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="MonacoClarionSourceEditor.cs" />
+    <Compile Include="Services\DocumentOutlineBuilder.cs" />
     <Compile Include="AgentPanelPad.cs" />
     <Compile Include="AgentPanelControl.cs" />
     <Compile Include="ShowAgentPanelCommand.cs" />
--- a/ClarionAssistant/MonacoClarionSourceEditor.cs
+++ b/ClarionAssistant/MonacoClarionSourceEditor.cs
@@ -589,6 +589,28 @@ namespace ClarionAssistant
             });
         }
 
+        // {action:"documentStructure"} — outline tree for the whole source file (this overlay always edits a
+        // real file 1:1, so LSP line 0-based -> Monaco 1-based is simple +1). Feeds the structure fly-out.
+        void IMonacoEditorHost.OnDocumentStructure(MonacoEditorControl editor, string rawJson)
+        {
+            int reqId, line, col; string buffer;
+            if (!ParseLspRequest(rawJson, out reqId, out line, out col, out buffer)) return;
+            System.Threading.Tasks.Task.Run(() =>
+            {
+                var symbols = new List<Dictionary<string, object>>();
+                try
+                {
+                    EnsureLsp();
+                    var resp = SharedLspBridge.GetDocumentSymbols(_filePath, buffer);
+                    object res = (resp != null && resp.ContainsKey("result")) ? resp["result"] : null;
+                    symbols = DocumentOutlineBuilder.Build(res, lsp0 => lsp0 + 1);
+                }
+                catch (Exception ex) { MonacoSpikeLog.Write("overlay documentStructure error: " + ex.Message); }
+                try { editor.PostResponse(reqId, new Dictionary<string, object> { { "symbols", symbols }, { "fileMode", true } }); }
+                catch { }
+            });
+        }
+
         /// <summary>Capture the active Clarion IDE theme's toolbar gradient (SerenityBlue/OfficeXP/Win10Blue/…)
         /// into _chromeBg1/_chromeBg2/_chromeFg so our overlay toolbar chrome follows it — the same colors the
         /// embeditor reads off its native ToolStrip, but sourced from the global ToolStripManager.Renderer (the IDE
--- a/ClarionAssistant/Services/DocumentOutlineBuilder.cs
+++ b/ClarionAssistant/Services/DocumentOutlineBuilder.cs
@@ -1,0 +1,87 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace ClarionAssistant.Services
+{
+    /// <summary>
+    /// Turns an LSP textDocument/documentSymbol result (hierarchical DocumentSymbol[] OR flat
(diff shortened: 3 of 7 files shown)


ClarionLive:
Integrated into master via 2d41931 (same route as #104: the branch was stacked on `feat/embeditor-in-editor-find`, which landed with additional find-switch fixes, so GitHub couldn't merge this cleanly).

What landed:
- `f6f0c7f` + `7a8755e` cherry-picked as-is (c155acf, b3c1181).
- `b847a14..e24e5a0` squashed into 9391fa8, with the outline block relocated next to the canonical `setEditorInsets` and the pad/overlay find-mode dispatch (#66 phase 2) kept — Ctrl+F/Ctrl+Shift+F route through `openFindUi`/`postCaFindOpen`, and the outline's left-column mutual exclusion with Find-All is preserved on both sides.
- Bonus fix: the #66 pad integration had dropped `applyFindFont`/`bumpFindFont` but left the find panel's Ctrl+wheel handler calling `bumpFindFont` (ReferenceError). Restored with this integration.

Verified: outline content byte-identical to this branch (relocation only), `node --check` clean on all inline scripts, Debug build green as 5.3.775. Thanks Mark — the outline + Class ▸ Methods regroup is a great addition, and the v1.4.1 addin reflection story made the bridge side painless.

Issue #113, opened 2026-07-17
feat(embeditor): open Find-All results in an editor tab — context, click-through, and filtering


msarson:
## Summary

Add VS Code's "open search results in editor" to CA Find: a **Find-All results document** listing every match with surrounding context and click-through back to the source, opened from **either** find surface (the overlay and the CA Find pad).

Follows on from #66 / #104 — that gave us the match engine and two front-ends; this gives the all-matches view a home that can actually hold it.

## Why a tab and not a pad

Three reasons, in order of weight:

1. **It's immune to the docking bug.** The IDE's vendored WeifenLuo DockPanel Suite v1 misplaces its dock guides whenever a monitor sits above or left of the primary (filed with SoftVelocity; not fixable from an addin). Anything pad-shaped inherits that. An editor tab never touches WinForms docking.
2. **The addin already owns five pads** and five `Ctrl+Alt` accelerators. This adds none.
3. **The all-matches grid is the part of find that genuinely wants persistent space** — which is a document's job. Quick find stays in the overlay where it belongs.

## What it does

* **⇗ button** in the Find-All header (overlay) and the results toolbar (pad) — both route through the *same* engine function, so the tab is identical from either surface and always lands on the right editor.
* Opens a `Search: <query>` tab: every match with **2 lines of context** either side, read-only (generated) hits marked, and the match span highlighted.
* **Click-through** reveals in the origin editor **and raises its tab**.
* **Filtering** over the loaded results (inspired by KSS/Open's Find-and-Delete, but non-destructive — rows are hidden and counted, never deleted):
  * scope chips with live counts — `All (329)` / `Editable (4)` / `Read-only (325)`
  * filter box with match-case, regex, and **invert** (KSS's "delete rows when they do NOT match")
  * `showing N of M` whenever a filter is active

The scope chips matter more than they sound in PWEE: searching `Long` in a NetTalk-heavy procedure gives **329 matches, 325 of them read-only** — nearly all `PROCEDURE(long …)` in generated DERIVED prototypes. One click gets you to the four that are yours.

## Design notes

**No stored positions.** Rows carry only the engine's array index; a click forwards `{op:'reveal', idx}` to the origin page, which resolves it against its live Monaco decorations (`syncFindMatches`, #65). That's exactly the pad's existing contract, so buffer edits can't make the tab navigate to the wrong line.

[...]

PR #117, opened 2026-07-17
fix(lsp): revert the module shadow when an overlay-mode embeditor closes (#115)

### LINKED ISSUE #115: LSP shadow of the generated module is never reverted when an overlay-mode embeditor closes
## Summary

`EmbedLspContext.RevertShadow()` has exactly one call site — `ModernEmbeditorViewContent.Dispose()`. An **overlay-mode** embed session never reaches it, so the LSP keeps a stale shadow of that generated module for the rest of the session.

## Why overlay mode never gets there

`Dispose()` is `AbstractViewContent.Dispose` — the workbench calls it when a **view** closes. An overlay-mode embeditor is never `ShowView`'d: it's a Monaco panel docked *over* the native PWEE editor (`ShowAsEmbedOverlay`). The workbench doesn't own it and never disposes it. `DetachOverlay()` is the entire teardown, hooked off `ClaGenEditor.Disposed`.

Comparing the two paths:

| | `Dispose()` | `DetachOverlay()` |
|---|---|---|
| `CaFindBroker.UnregisterHost` | ✔ | ✖ (being fixed in #113/#114) |
| **`_lspContext.RevertShadow()`** | ✔ | **✖ — this issue** |
| `_instances.Remove(this)` | ✔ | ✔ |
| `_settingsReg.Dispose()` | ✔ | ✔ |
| `_panel.Dispose()` | ✔ | ✔ |

## Impact

From `EmbedLspContext` (#56 / PR #74): while an embeditor is open, every LSP request pushes the wrapped embed buffer to the server under `RealPath`, **overriding the on-disk file in the server's view**. The transport has no `didClose`, so `RevertShadow` — pushing the real on-disk content back — is the only thing that un-shadows it.

Skip it and, after any overlay-mode PWEE close, the language server still believes that generated module *is* the embed buffer. For the rest of the session:

* hover / go-to-definition / find-references on that `.clw` answer from a buffer that is no longer open;
* the staleness is **silent** — results look authoritative, they're just from the wrong content;
* it persists until the LSP restarts, and compounds per procedure closed.

This undermines #56, whose whole point is that the LSP sees the *right* module content.

## Fix

Call `RevertShadow()` from `DetachOverlay()`, mirroring `Dispose()`. Same best-effort/try-catch posture — it's already a safe no-op when the LSP isn't running or the file has vanished.

## Note

Overlay mode never receiving `Dispose()` is a **pattern**, not a one-off — anything placed in `Dispose()` is silently skipped for every overlay embed session. `UnregisterHost` was the first casualty (found via #113, fixed in #114); this is the second. Worth a look at whether anything else should be reachable from both teardown paths.

msarson:
Closes #115. One file, twelve lines, branched from master.

`EmbedLspContext.RevertShadow()` had exactly one call site — `Dispose()`. An **overlay-mode** embed session never reaches it, because it is never `ShowView`'d: the workbench doesn't own the view and never disposes it. `DetachOverlay()` is the entire teardown.

## Why it matters

From #56 / PR #74: while an embeditor is open, every LSP request pushes the wrapped embed buffer to the server under the module's **real path**, overriding the on-disk file in the server's view. The transport has no `didClose`, so `RevertShadow` — pushing the on-disk content back — is the only thing that un-shadows it.

Skip it and, after any overlay-mode PWEE close, the server still believes that generated module *is* the embed buffer. Hover, go-to-definition and find-references on that `.clw` keep answering from a buffer that's no longer open — **silently**, since the results look authoritative — for the rest of the session, compounding per procedure closed. That undermines the point of #56.

## The fix

Call `RevertShadow()` from `DetachOverlay()`, mirroring `Dispose()` and its best-effort try/catch. It's already a safe no-op when the LSP isn't running or the file has vanished.

## Integration note

This touches `DetachOverlay()`, which **#114 also touches** — adding the `CaFindBroker.UnregisterHost` that's missing from the same path for the same underlying reason. Expect a trivial conflict at the same anchor; the two additions are independent and both belong there. When both land the order matches `Dispose()`: unregister, then revert.

Deliberately **not** stacked on #114 — a two-line conflict is cheaper to resolve than an unmergeable stack (cf. #110).

## The wider point

Overlay mode never receiving `Dispose()` is a **pattern**, not a one-off: anything placed in `Dispose()` is silently skipped for every overlay embed session. `UnregisterHost` was the first casualty (found via #113, fixed in #114); this is the second. The two paths are now aligned, but it's worth deciding whether they should share one teardown rather than being kept in sync by hand.

Built clean against Clarion 11.1. Not runtime-verified — the failure is silent by nature (stale LSP answers), so I'd want an LSP-level check rather than claiming I watched it work.

### DIFF
--- a/ClarionAssistant/Terminal/ModernEmbeditorViewContent.cs
+++ b/ClarionAssistant/Terminal/ModernEmbeditorViewContent.cs
@@ -2248,6 +2248,18 @@ namespace ClarionAssistant.Terminal
             if (_overlayDetached) return;
             _overlayDetached = true;
             UnhookOverlayTeardown();
+
+            // #56: while this embed session was up, every LSP request pushed the wrapped embed buffer to the
+            // server under the module's REAL path, overriding the on-disk file in the server's view. There is
+            // no didClose in the transport, so RevertShadow (pushing the on-disk content back) is the only
+            // thing that un-shadows it.
+            //
+            // Dispose() already does this — but overlay mode never GETS Dispose(): it is never ShowView'd, so
+            // the workbench does not own this view and never disposes it. DetachOverlay IS the teardown. Without
+            // this call the server keeps answering hover/definition/references for that .clw from a buffer that
+            // is no longer open, silently, for the rest of the session.
+            try { if (_lspContext != null) _lspContext.RevertShadow(); } catch { }
+
             RestoreNativeChrome();
             RemoveOverlayCover();
             try


ClarionLive:
Integrated into master via 53816a9 — the predicted two-line conflict with #114 in `DetachOverlay()` resolved exactly as you specified: `UnregisterHost`, then `RevertShadow`, matching `Dispose()` order. The resolved tree was build-verified together with #111/#112/#114 before anything was merged.

Agreed on the wider point: `Dispose()` and `DetachOverlay()` being kept in sync by hand is now a two-casualty pattern — worth an issue for a shared teardown. Closing as merged (the conflict meant GitHub couldn't take it directly).

PR #129, opened 2026-07-18
fix(editor): readable toolbar text on dark IDE chrome + read-only shading under Windows High Contrast

ClarionLive:
## The two reports (screenshots in chat)

1. **Dark Clarion IDE:** the CA Editor's `Reload` and `Save` buttons (and the title) are unreadable — dark text on the dark captured chrome gradient.
2. **Windows High Contrast:** the CA Embeditor's read-only (generated) section shading is invisible.

## Root causes & fixes

**1. No contrast guarantee between captured chrome bg and fg.** The toolbar inherits the IDE theme's gradient (`--chrome-bg1/2`), but the text color is unreliable: the source editor sends no fg at all (`ProfessionalColorTable` carries no text color — the page then falls back to its *own* theme's color, which tracks the page pref, not the IDE gradient), and the embeditor's captured `strip.ForeColor` can be ControlText black even on a dark IDE theme. Fix: a contrast guard in the chrome-apply handler — compute the gradient's relative luminance (`hexLum`); when the supplied fg is missing or within 0.3 luminance of the gradient, substitute `#e8e8f0` (dark chrome) / `#1a1a24` (light chrome). A well-contrasted host-supplied fg passes through untouched.

**2. Forced-colors strips subtle alpha washes.** "High Contrast mode" here is Windows HC (there is no CA toggle by that name) — under `forced-colors`, the 5%-alpha read-only washes are stripped or imperceptible on the forced canvas. Fix: `@media (forced-colors: active)` opts `.embed-readonly-line` out of color forcing, with a stronger neutral wash (`rgba(127,127,127,.28)`) plus an inset left-edge bar — mid-gray reads on both black- and white-based HC themes.

## Verification

- Luminance guard unit-checked for all four cases: dark-gradient+missing fg → light substitute; light-gradient+missing → dark substitute; black-on-navy → substituted; black-on-light → kept.
- NUL-byte scan clean; all inline script blocks pass `node --check`; Debug build clean.
- In-IDE pass next deploy: dark IDE theme → Reload/Save/title readable on the chrome gradient; toggle Windows HC → generated lines visibly shaded with the left-edge bar.

### DIFF
--- a/ClarionAssistant/Terminal/monaco-embeditor.html
+++ b/ClarionAssistant/Terminal/monaco-embeditor.html
@@ -299,10 +299,17 @@
        is retained (invisible) so the read-only guard can query live decoration ranges. */
     .embed-readonly-line { background: rgba(0, 0, 0, 0.06); }
     body:not(.light) .embed-readonly-line { background: rgba(255, 255, 255, 0.05); }
-    /* Windows High Contrast (forced-colors): the subtle alpha washes above are stripped/imperceptible on
-       the forced canvas, so read-only (generated) lines became indistinguishable from editable ones
-       (John's report). Opt the decoration out of color forcing and use a stronger neutral wash plus a
-       left edge cue — mid-gray reads on both black- and white-based HC themes. */
+    /* Monaco's built-in "Toggle High Contrast Theme" (editor context menu) applies theme class hc-black
+       (hc-light on newer builds) to the editor root — the pure-black canvas makes the 5%-alpha washes
+       above imperceptible, so read-only (generated) lines became indistinguishable from editable ones
+       (John's report, screenshot 2). Stronger neutral wash + a left edge cue; mid-gray reads on both. */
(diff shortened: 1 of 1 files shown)


ClarionLive:
Updated after John identified the toggle: it's **Monaco's own** context-menu action ("Toggle High Contrast Theme"), not Windows HC — and the answer to "is it saved somewhere?" is **no**: Monaco flips the theme in memory only, nothing is written anywhere, so there was nothing to read back.

Second commit adds:
- **HC persistence**: the page watches the editor root's theme class (`hc-black`/`hc-light`), stores the state in localStorage next to the dark pref, and restores on boot by triggering Monaco's own toggle action (version-proof — no theme-name guessing). The light/dark pill exits HC by design and the observer persists that exit.
- **hc-theme-scoped read-only shading** (the actual repro in screenshot 2): same stronger wash + left-edge bar as the forced-colors rule, which stays for genuine OS High Contrast users.

PR #144, opened 2026-07-23
fix(editor): mirror native caret jumps in Monaco (Errors pane selections did not go to correct posision in file)

geircodes:
## Summary

Clicking a line in the Clarion IDE's native Errors pane (compiler error/warning list) opened the correct file in the Monaco/WebView2 overlay, but the caret landed on the wrong line — sometimes line 1 on a fresh open, sometimes a stale unrelated line left over from a previous session (observed: error reported at line 1884, caret landed at line 1944).

## Root cause

`MonacoClarionEditor` only ever learns about a navigation target when the caller goes through `MonacoSourceNavigator.NavigateToFileAndLine` (used by our own MCP `open_file` tool and by other addins like the standalone debugger). Two-way caret sync between the hidden native editor and the visible Monaco surface was never implemented — the class's own comment says so directly ("Caret/document two-way sync is Phase 2 — for now we push the buffer once on load").

The Errors pane (like Bookmarks next/prev, Find Next, and the breakpoint list) is native, closed-source IDE functionality — it moves the hidden `ICSharpCode.TextEditor.Caret` directly, the same way `EditorService.GoToLine`/`NavigateToFileAndLine` do, and never touches `MonacoSourceNavigator`. With no live sync, Monaco simply keeps showing whatever it last had.

## Fix

`ICSharpCode.TextEditor.Caret` exposes a public `PositionChanged` event. `MonacoClarionEditor` now subscribes to it on the captured native `ActiveTextAreaControl.Caret` once the overlay attaches, and mirrors any external line change into the visible Monaco surface via the existing `NavigateToLine` path (which only posts a WebView2 `revealLine` message — it never writes back to the native `Document`/`Caret`, so there's no feedback-loop risk).

This is a general fix, not an Errors-pane-specific patch: it covers ANY native code path that moves the caret directly, since it listens on the caret itself rather than chasing individual call sites.

Guards:
- Line-delta check (`_lastMirroredCaretLine`) — `GoToLine` sets `Caret.Line` then `Caret.Column` as two separate writes, firing `PositionChanged` twice per native jump; the delta guard coalesces that into a single mirror.
- Defensive UI-thread marshalling via `WorkbenchSingleton.Workbench as Form`, same pattern already used in `MonacoSourceNavigator.NavigateToFileAndLine`, since `PositionChanged`'s firing thread isn't guaranteed across every native caller.
- Unsubscribed in `Dispose()` alongside the existing `MonacoSourceNavigator.Unregister` call — no new teardown path.

## Verification

Live-tested in the running Clarion IDE: clicking an Errors-pane entry now lands Monaco's caret on the correct line — and, as a bonus of Monaco's own column-aware reveal, at the actual variable/function token the error refers to, not just the right line. Tested against both an already-open file and a cold file opened for the first time (the two distinct symptom variants originally reported).

An independent review pass (fresh agent, no context carried over from drafting) confirmed the root-cause diagnosis, found no existing caret-event subscription anywhere in the codebase, verified the feedback-loop analysis (no Monaco→native write path exists today), and confirmed the line-delta + UI-thread-marshal design before implementation.

### DIFF
--- a/ClarionAssistant/MonacoClarionSourceEditor.cs
+++ b/ClarionAssistant/MonacoClarionSourceEditor.cs
@@ -48,6 +48,8 @@ namespace ClarionAssistant
         private Timer _coverSafety;      // backstop: drop the cover even if the page never signals nav-completed
         private Control _navBar;         // native QuickClassBrowser bar (class/members combos) hidden while the overlay covers the editor; restored on teardown
         private ICSharpCode.TextEditor.TextEditorControl _hostEditor;   // the live editor we mirror
+        private ICSharpCode.TextEditor.Caret _watchedCaret;   // native caret mirrored into Monaco while overlay is on
+        private int _lastMirroredCaretLine = -1;              // 0-based; guards PositionChanged's double-fire (GoToLine sets Line then Column separately)
         private string _overlayTitle = "Clarion Source";
         // Active Clarion IDE theme's toolbar gradient (SerenityBlue/OfficeXP/…) → our overlay toolbar chrome, so
         // the CA Editor's toolbar matches the IDE theme like the embeditor's does. Captured at OnReady. (task c8e669d3)
@@ -246,6 +248,7 @@ namespace ClarionAssistant
                 _settingsReg = Services.MonacoSettingsBroadcaster.Register(json => { try { _editor?.PostJson(json); } catch { } });
                 if (_cover != null) _cover.BringToFront();   // cover ABOVE the WebView2 until Monaco has painted
                 WireBreakpoints();   // keep Monaco's gutter in sync with IDE breakpoints
+                WireCaretSync(host);   // mirror native caret jumps (Errors pane, Bookmarks, Find Next, breakpoint list) into Monaco
 
                 // Backstop: reveal Monaco after a few seconds even if OnEditorNavigationCompleted never arrives,
                 // so a failed load can never leave the cover stranded over a blank editor.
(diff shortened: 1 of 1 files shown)


peterparker57:
Follow-up after live testing (John's multi-error repro — 7 errors + 6 warnings in one generated .clw): the caret-event mirror turned out to have two blind spots, now fixed on master in three parts. Sharing since it's a useful map of how Errors-pane navigation actually routes:

1. **857d92d** — the Errors pane goes `FileService.JumpToFilePosition → IPositionable.JumpTo`, and the native caret setter fires **no PositionChanged at all** when it's already at the target — so a click could be a complete no-op depending on where the hidden native caret happened to sit. `JumpTo` is a sealed interface impl on the wrapper (couldn't override), so `MonacoClarionEditor` now re-implements `IPositionable` and mirrors every jump unconditionally with the exact requested position. Your PositionChanged mirror stays for the movers that bypass JumpTo (bookmarks, Find Next).
2. **ac27e26** — some error rows don't touch the source document at all: Clarion routes them through its own error→embed mapping into the PWEE embeditor. The CA overlay only sampled that caret once at attach (pre-paint, before Clarion positioned it), and later clicks into the already-open embed moved only the hidden native caret. The overlay now subscribes to the embed caret's PositionChanged — same technique as yours, applied to the embeditor.
3. **1312f3c** — nastier consequence of that same route: Clarion closes + re-opens the native embed when navigation crosses between source-routed and embed-routed rows, which tore down a dirty overlay silently (the native buffer is clean, so Clarion never prompts). The page now mirrors its edited slots to the host per edit; an external teardown stashes them, and the next attach for the same procedure restores them against a byte-identical baseline.

All three verified live. Thanks for the original mirror — it was the right instinct, the fork just hides more navigation routes than any one hook can see.

PR #146, opened 2026-07-23
feat(source-editor): watch for external readonly/content changes on disk

geircodes:
## Summary

The CA Editor had no way to notice a file changing outside the IDE while its tab was open — neither its Windows read-only attribute flipping (e.g. a SourceSafe checkout/checkin) nor its content changing (another editor, another tool). The developer would keep editing a stale buffer, or keep believing a file was still writable/read-only after it had actually changed, with no indication until they happened to close and reopen the tab.

## Fix

Each CA Editor tab now runs its own `FileSystemWatcher` (wired in `OnReady`, torn down in `Dispose`) on the file it's showing, debounced ~350ms before re-stat'ing (SourceSafe/antivirus/atomic-replace saves often fire several raw filesystem events per real change).

- **Readonly ↔ readwrite only, content unchanged:** applied silently — Monaco's `readOnly` option, the Save button, and the title's `(read-only)` suffix update with no popup. This was the original, explicit ask: no user interaction for this case.

- **Content changed on disk:** a toast with **Reload** and **Show diff** (disk vs. the live editor buffer — including any unsaved edits — via `MonacoDiffViewContent`, the same diff surface `show_diff` uses). If there are unsaved local edits, the toast text says so.<br/>
&emsp;&emsp;Toast_FileChanged
&emsp;&emsp;Toast_BothPartiedModifiedFile

- **File deleted/moved externally:** a toast with **Close tab** (closes via the same cancellable `ClosingEvent` path a normal tab-close uses, so an unsaved-edits prompt still fires; Save still recreates the file either way).<br/>
&emsp;&emsp;Toast_FileDeleted

- The toast never auto-dismisses on its own — silently losing the notice would leave a stale tab with no visible sign anything changed. A manual **✕** dismiss and **Escape** are provided instead, since neither Reload/Close tab is always the right immediate choice.
- Self-writes (interactive save, the existing build-triggered save, Reload) refresh an internal baseline immediately after writing, so they're never mistaken for an external change — even a watcher event for our own write arriving late will re-stat and find nothing different from the baseline. No suppress-flag/timing-window needed.
- The watcher observes the whole containing directory (`FileSystemWatcher.Filter` is applied client-side, not by the OS), so unrelated heavy churn there could overflow its internal buffer; on that `Error` it disposes and recreates itself rather than silently going dark.
- The two new page→host actions (`diffWithDisk`, `closeTab`) are routed through the existing `OnUnknownAction` catch-all (same mechanism `toggleBreakpoint` already uses) rather than adding new `IMonacoEditorHost` methods — so `ModernEmbeditorViewContent` (CA Embeditor, which implements the same shared interface) is completely untouched. This is scoped to the CA Editor (whole-file overlay) only.

Also fixed two dead buttons found while live-testing this:
- `MonacoDiffViewContent`'s Approve/Cancel had no handler wired for this comparison-only use — both now just close the diff tab.
- The toast's "Show diff" button used to dismiss the toast on click even though nothing was actually resolved (any button click dismissed). It now opts out via `dismiss:false`, so Reload/Close tab are still there after you've looked at the diff.

## Abandoned: background-tab title marker

Also attempted: prefixing the native tab's title (e.g. `⚠ filename.clw`) when a change notice fires while that tab isn't the active one, so it's visible from the tab strip without switching to it. Confirmed the underlying data change was correct (title string genuinely updated, correct repaint call invoked with no error) but the visible tab label itself never updated in testing, across several different approaches. Dropped it rather than keep guessing at what turned out to be a deeper rendering question than expected — the toast itself is unconditional and reliable regardless, so the substantive notification isn't lost, just this one visual enhancement on top of it. If you've run into something like this before, or have a pointer, that'd be very welcome — happy to pick it back up.

## Verification

Live-tested extensively against a real file outside the project (edited/saved from Notepad++, deleted, recreated via Save) covering: readonly flip while open (silent, both directions), content change with and without local unsaved edits, delete + Close tab + unsaved-edit prompt, Save recreating a deleted file without falsely re-triggering the watcher, Reload from the toolbar correctly failing on a genuinely-deleted file, and self-writes (interactive save, Reload) never producing a false external-change notice. An independent code-reviewer pass caught and fixed a real post-`Dispose` resurrection race in the watcher/debounce-timer teardown before this was finalized.

### DIFF
--- a/ClarionAssistant/MonacoClarionSourceEditor.cs
+++ b/ClarionAssistant/MonacoClarionSourceEditor.cs
@@ -57,6 +57,19 @@ namespace ClarionAssistant
         private string _overlayLiveText; // last live buffer the page mirrored, for the close-save write
         private IDisposable _settingsReg; // registration in MonacoSettingsBroadcaster (cross-surface gear-settings sync, deac3d16)
 
+        // External disk watch: SourceSafe checkout/checkin flips the Windows read-only attribute, and any
+        // outside editor (Notepad++, another IDE) can change the content — neither is visible to us without
+        // this. See WireDiskWatch/CheckDiskState below for the full mechanism.
+        // _diskWatchDisposed guards a real race: a watcher/Error event can already be queued via BeginInvoke
+        // onto the (long-lived) workbench form BEFORE Dispose runs, and execute AFTER it — without this flag,
+        // that queued callback would recreate a live FileSystemWatcher/Timer that nothing will ever tear down.
+        private FileSystemWatcher _diskWatcher;
+        private Timer _diskWatchDebounce;
+        private volatile bool _diskWatchDisposed;
+        private bool _lastKnownReadOnly;
+        private DateTime _lastKnownWriteUtc;
+        private long _lastKnownLength = -1;
+
         // Host-driven navigation (debugger / breakpoint-list click via MonacoSourceNavigator). The page can't
         // be positioned until it signals ready, so a nav that arrives earlier is parked here and flushed in
         // OnReady. 1-based; _navPendingLine == 0 means "nothing queued".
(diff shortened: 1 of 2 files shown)


geircodes:
The last commit adds a toast like this when a edited file becomes read only by an external action 

image

This makes it impossible to lose edits without knowing. Closing the file will give this dialog: 

image

So either you cancel and reenable the read/write and save the file, or press OK and the edits are discarded. The toast "diff" button will let you see the actual changes.

PR #165, opened 2026-07-31
fix(codegraph): resolve caller attribution for overloaded procedures/methods

geircodes:
## Summary

CodeGraph's "what does this procedure/method call" resolution silently collapses every overloaded
Clarion procedure or class method (several declarations sharing the exact same name, differing only
by parameter type — a legal, common Clarion pattern) onto a single arbitrary overload. Every call
and variable reference made from *any* of the overloads' bodies gets attributed to whichever overload
happened to be loaded last, regardless of which overload's body the scanner was actually inside.

Confirmed against a real production Clarion solution: **65 distinct overloaded names** (2 to 6 overloads each,
145 overload symbols total), with **4,372 `calls`/`references` relationship rows** collapsed onto
one overload per name. Only 66 of those 145 overload symbols ever appeared as a caller of anything
in any query — the other 79 were completely invisible; asking "what does this specific overload
call?" for any of them silently returned nothing, or someone else's calls. After the fix: 139 of 145
now correctly appear as callers (the remaining 6 have genuinely empty bodies).

## Root cause

`ResolveRelationships` in `CodeGraphIndexer.cs` tracks "which procedure is the current source line
inside" (`currentProcId`) via dictionaries keyed by **name only**:

- `symbolByFile` / the per-file `currentFileSymbols` lookup (`filePath → name → id`)
- `symbolNameToId` (global `name → id`, explicitly documented as "last wins — implementation in
  member file overwrites MAP declaration")

Both can hold exactly one id per name. That's fine for the one case the "last wins" comment was
written for (a MAP forward-declaration and its later real implementation, both surfacing under the
same name). It silently breaks for genuine overloading: several *simultaneously real* procedures
sharing one name each get their own row in the `symbols` table (correct — the symbol/parsing side was
never the problem), but the relationship-resolution pass's `currentProcId = currentFileSymbols[matchedName]`
lookup can only ever return one of them — the one every OTHER overload's calls/references also get
attributed to, incorrectly.

Four other places in the same method **reverse-scan** one of these name-keyed dictionaries looking
for "the entry whose value equals `currentProcId`" (to recover the current procedure's *name*, e.g.
to derive its owning class for `SELF.Method` resolution, or to scope a parameter lookup to the right
procedure). These are independently broken by the same root cause: even fixing the id-side lookup
above doesn't fix these, since a reverse scan over a name-keyed dict still can't distinguish which of
several overloads sharing one dictionary entry a given id actually belongs to.

## Fix

Added a new per-file `(definition line → id)` lookup (`symbolLineByFile` / `currentFileSymbolsByLine`),
populated from the `symbols` table's existing `line_number` column (only the `SELECT` needed
widening). `procDefRegex` only ever matches a line that *is* a procedure's own definition line, so
resolving `currentProcId` by `(file, exact line)` instead of `(file, name)` is unambiguous regardless
of how many overloads share a name — no heuristics, no signature parsing needed for this half of the
problem.

Introduced a `currentProcName` string, updated in lockstep with `currentProcId` at every one of its
three assignment sites (initialization from `parentProcId`, the local-MAP reset case, and the main
per-line resolution case), and used it to replace all four reverse-scan loops directly — since the
matched procedure's name text is already in hand at the exact moment `currentProcId` is resolved,
no scan is needed at all.

Applied identically to both synced copies (`CodeGraph/Graph/CodeGraphIndexer.cs`, the in-process
copy, and `indexer/Graph/CodeGraphIndexer.cs`, the standalone CLI copy).

**Independent code-reviewer agent verification done before applying** (per this project's standard
practice): walked every `currentProcId` assignment site to confirm `currentProcName` is correctly
paired at all of them (one initially-suspected gap was checked directly against the code and
retracted — no missing pairing found), confirmed no 5th reverse-scan site was missed, confirmed the
two synced copies are substantively identical, confirmed the 1-based `line_number` column vs.
0-based loop-index (`i`/`p`) convention is applied consistently at both write and read time, and
confirmed no regression risk for the common non-overloaded case (a real implementation's definition
line always has a `symbolLineByFile` entry, since the same regex/line convention feeds both the
symbol table and this resolver).

## Deliberately out of scope

This fix only corrects the **caller** side — "what does this specific overload call" — not the
**callee** side: which overload a given *call site* actually targets is still resolved by the old
name-keyed `symbolNameToId` lookup, so it still picks whichever overload loaded last, same as
before. Real disambiguation there needs type-aware overload resolution (matching call-site argument
types against each candidate's declared parameter types) — a materially bigger feature, not
attempted here. Per-overload *caller counts* (grouping by call-target id) remain unreliable for an
overloaded name; "what does this overload call" (grouping by caller id) is now correct. See the
fixture's Bug P writeup for the exact before/after boundary, including a documented case
(`SELF.Dispatch(...)`) where the call still resolves to the wrong sibling overload on purpose, to
make clear that's not a regression.

A related, pre-existing, NOT-fixed-by-this-PR gap noticed during verification: local-variable
*references* can still cross-contaminate between overloads that happen to share a variable name,
since `VariableInfo.ParentName` is stored as plain procedure-name text (identical across all
overloads of that name), not tied to a specific overload's symbol/line. Unaffected by this fix
either way — not introduced, not worsened, not fixed.

## Minimal reproduction

Extended the shared `test-fixtures/codegraph-repro/` fixture: `OverloadBugClass.Dispatch` is
declared twice, differing only by parameter type (`LONG` vs `*CSTRING`). The `*CSTRING` overload
delegates to the `LONG` overload via `SELF.Dispatch(...)` — the same general shape as the
motivating case (an overload calling a sibling overload of the same name). Each overload calls a
different, already-existing target (`WorkerClass.Sign` from the `LONG` overload, `WorkerClass.Ask`
from the `*CSTRING` overload) so the two bodies' own calls can be told apart in the results:

```clarion
OverloadBugClass.Dispatch PROCEDURE( LONG pValue )
worker WorkerClass
  CODE
  RETURN worker.Sign( pValue )

OverloadBugClass.Dispatch PROCEDURE( *CSTRING pValue )
result LONG
worker WorkerClass
  CODE
  result = SELF.Dispatch( 99 )
  result = worker.Ask( result )
  RETURN result
```

**Before this fix**: both `worker.Sign(pValue)` (inside the `LONG` overload) and
`worker.Ask(result)` (inside the `*CSTRING` overload) would be attributed as coming from the same
one overload symbol — whichever loaded last — never from their own respective overloads.

**After this fix** (verified via the standalone `clarion-indexer.exe`): `WorkerClass.Sign` gains
exactly one caller (21 → 22), correctly attributed to the `LONG` overload's own definition line.
`WorkerClass.Ask` gains exactly one caller (2 → 3), correctly attributed to the `*CSTRING`
overload's own definition line — a different id than the `Sign` caller above, despite both sharing
the literal name `OverloadBugClass.Dispatch`. Zero regressions on any prior fixture case.

**Also verified against the real production solution**: reindexing confirmed the from-side of every
one of the 65 overloaded names now correctly distributes its relationships across its own overloads
instead of collapsing onto one (see Summary for the aggregate numbers).

## Scope note

Branches cleanly from `upstream/master` (built and verified in an isolated worktree, no dependency
on any other unmerged branch).

### DIFF
--- a/ClarionAssistant/CodeGraph/Graph/CodeGraphIndexer.cs
+++ b/ClarionAssistant/CodeGraph/Graph/CodeGraphIndexer.cs
@@ -563,16 +563,29 @@ namespace ClarionCodeGraph.Graph
             var procNames = new List<string>(); // ordered list for matching
             // File-specific lookup: filePath → (name → id) — resolves ambiguous names
             var symbolByFile = new Dictionary<string, Dictionary<string, long>>(StringComparer.OrdinalIgnoreCase);
+            // File-specific lookup: filePath → (definition line → id). Clarion allows several
+            // procedures/methods to share the exact same name via parameter-type overloading
+            // (e.g. a method declared once per parameter type) -- symbolByFile and
+            // symbolNameToId can only ever hold ONE id per name, so they silently collapse every
+            // overload onto whichever one happened to be inserted last (see the "Last wins"
+            // comment below). A procedure's own definition line is always unique per file, so
+            // it's used below to pick out the exact overload currentProcId/currentProcName
+            // should track while scanning that overload's own body.
+            var symbolLineByFile = new Dictionary<string, Dictionary<int, long>>(StringComparer.OrdinalIgnoreCase);
 
             var allSymDt = _db.ExecuteQuery(
-                "SELECT id, name, type, file_path FROM symbols WHERE type IN ('procedure','function','routine')");
+                "SELECT id, name, type, file_path, line_number FROM symbols WHERE type IN ('procedure','function','routine')");
 
             foreach (System.Data.DataRow row in allSymDt.Rows)
             {
                 string name = row["name"].ToString();
                 long id = Convert.ToInt64(row["id"]);
                 string filePath = row["file_path"].ToString();
-                // Last wins — implementation in member file overwrites MAP declaration
+                int lineNumber = row["line_number"] != DBNull.Value ? Convert.ToInt32(row["line_number"]) : -1;
+                // Last wins — implementation in member file overwrites MAP declaration. Also the
+                // known limitation this fix works around: for genuine overloads sharing a name,
+                // only the last-loaded one survives here. symbolLineByFile above is the
+                // disambiguated path for anything that needs the correct per-overload id.
                 symbolNameToId[name] = id;
 
                 // Build per-file symbol lookup
@@ -583,6 +596,17 @@ namespace ClarionCodeGraph.Graph
                     symbolByFile[filePath] = fileSymbols;
                 }
                 fileSymbols[name] = id;
+
+                if (lineNumber > 0)
+                {
+                    Dictionary<int, long> fileSymbolsByLine;
+                    if (!symbolLineByFile.TryGetValue(filePath, out fileSymbolsByLine))
+                    {
+                        fileSymbolsByLine = new Dictionary<int, long>();
+                        symbolLineByFile[filePath] = fileSymbolsByLine;
+                    }
+                    fileSymbolsByLine[lineNumber] = id;
+                }
 
                 // Only add procedures/functions from .clw files to the match list.
                 // Skip: routines, dotted names (class method implementations),
@@ -757,13 +781,18 @@ namespace ClarionCodeGraph.Graph
                     int localClassEndDepth = 0;
                     // The parent (first) procedure in each member file owns all calls
                     long parentProcId = -1;
+                    string parentProcName = null;
                     // Track local MAP procedure names — skip these as call targets
                     var localMapNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
-                    // Get file-specific symbol lookup for this file
-                    Dictionary<string, long> currentFileSymbols;
-                    if (!symbolByFile.TryGetValue(target.Path, out currentFileSymbols))
-                        currentFileSymbols = new Dictionary<string, long>(StringComparer.OrdinalIgnoreCase);
+                    // Get file-specific (definition line → id) lookup. Used (instead of a
+                    // name-keyed lookup) everywhere the exact overload matters, since
+                    // symbolByFile/symbolNameToId can only hold one id per name and would
+                    // collapse same-named overloads onto whichever loaded last (see
+                    // symbolLineByFile's comment above).
+                    Dictionary<int, long> currentFileSymbolsByLine;
+                    if (!symbolLineByFile.TryGetValue(target.Path, out currentFileSymbolsByLine))
+                        currentFileSymbolsByLine = new Dictionary<int, long>();
 
                     // Load variables for this file
                     List<VariableInfo> currentFileVars;
@@ -783,11 +812,14 @@ namespace ClarionCodeGraph.Graph
                             {
                                 foundFirstProc = true;
                                 string matchName = firstMatch.Groups[1].Value;
-                                // Use ONLY file-specific lookup to avoid cross-file name collisions.
-                                // symbolNameToId may return a symbol from a different file if names collide.
+                                // Resolve by (file, definition line), not name alone -- see the
+                                // overload note where currentProcId is updated the same way below.
                                 long id;
-                                if (currentFileSymbols.TryGetValue(matchName, out id))
+                                if (currentFileSymbolsByLine.TryGetValue(p + 1, out id))
+                                {
                                     parentProcId = id;
+                                    parentProcName = matchName;
+                                }
                             }
                             continue;
                         }
@@ -817,6 +849,7 @@ namespace ClarionCodeGraph.Graph
                     if (parentProcId < 0) continue;
 
                     long currentProcId = parentProcId;
+                    string currentProcName = parentProcName;
                     bool seenFirstCode = false;
 
                     for (int i = target.StartLine; i < lines.Length; i++)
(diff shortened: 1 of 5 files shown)


ClarionLive:
Reviewed and ran an empirical before/after to check the one thing I was unsure about. **Result: no regressions, and the PR undersells itself in two ways.** Details below.

## What I wanted to check

The fix swaps a name-keyed lookup for a line-keyed one:

```csharp
if (currentFileSymbolsByLine.TryGetValue(i + 1, out id))
{ currentProcId = id; currentProcName = matchedName; }
else
{ currentProcId = parentProcId; currentProcName = parentProcName; }
```

That makes the `else` branch reachable in strictly more cases. The old code fell back to `parentProcId` only when the name wasn't anywhere in the file; the new code falls back whenever no symbol exists at that *exact line*. Any line where the indexer's `procDefRegex.Match(trimmed)` fires but the parser recorded no symbol would newly dump the rest of that procedure's calls onto the file's first procedure — the misattribution the code's own comment warns about:

> `// mere prototype -- resets currentProcId to parentProcId (the file's`
> `// FIRST procedure), silently misattributing every call made for the rest`
> `// of the REAL enclosing procedure to that unrelated first procedure`

The PR's evidence measures overload *coverage*, which doesn't rule this out.

## Method

Two git worktrees (`master` @ `eb5e71a` vs this PR @ `07fcd7f`), built the standalone `clarion-indexer` in each, indexed the same production solution (`v8POSitive.sln` — 1,659 files, 31 projects, 333,685 symbols) to separate DBs.

Symbol IDs came out identical across both runs (0 mismatches, 0 rows unique to either side), so call sites could be matched exactly on `(type, file_path, line_number, to_id)` rather than compared only in aggregate.

## Result: every call site accounted for

| Fate | Count |
|---|---:|
| Unchanged | 242,730 |
| Reattributed | 16,440 |
| **Lost** | **0** |
| Gained | 19,774 |

Of the 16,440 reattributed: 0 moved *off* the first procedure, 16,122 moved between two non-first procedures (the overload fix proper), and 318 moved *onto* the file's first procedure — the exact signature I was looking for.

**All 318 are fixes, not regressions.** They cluster in 8 files, and in every case the caller name is identical or a case variant before→after. `PRMLabel039.clw` shows the mechanism:

```
PriceIt              PROCEDURE (Save_Recs Save_Rec,Byte p:Action,...)   ! line 5
...
PriceIt                      Routine                                     ! line 121
```

A local ROUTINE shadowing its enclosing PROCEDURE's name. Under `OrdinalIgnoreCase` last-wins, `currentFileSymbols["PriceIt"]` resolved to the **routine** (inserted later), so the procedure's whole body was credited to the routine — including line 67, one line after `CODE` and 54 lines before the routine begins. After this PR it goes to the procedure at line 5. Correct.

Same shape elsewhere: `PRM002a073.clw` has `VoidPO` procedure (line 5) + `VoidPO` routine (line 106); `PRMGlr047.clw` has three case-variant procedures — `Setup` (256), `SetUp` (260), `SetUp` (264).

So this fixes a second flavour of the same root cause that isn't mentioned in the writeup: **a local ROUTINE (or a case variant) colliding with its enclosing procedure's name**, not just parameter-type overloads.

## Aggregate deltas

| Metric | before | after | Δ |
|---|---:|---:|---:|
| symbols | 333,685 | 333,685 | **0** |
| proc/func/routine symbols | 109,780 | 109,780 | **0** |
| `calls` rows | 33,319 | 53,093 | **+19,774 (+59%)** |
| `references` rows | 225,851 | 225,851 | 0 (redistributed, not added) |
| distinct symbols appearing as a caller | 31,297 | 37,527 | +6,230 |
| procs with zero outgoing rows | 22,107 | 15,867 | **−6,240** |

Identical symbol counts independently confirm the claim that symbol capture was never the problem. References redistribute at constant total; calls *increase*, because a correct `currentProcName` lets `SELF.Method` / `ClassName.Method` resolution fire where it previously derived a wrong class prefix or null. 6,240 procedures that were completely invisible as callers now appear.

## Bonus: ~40% faster indexing

| | duration |
|---|---:|
| before | 1,770,761 ms (29.5 min) |
| after | 1,057,980 ms (17.6 min) |

That's the four O(symbols) reverse-scans this PR deletes from the per-line loop — e.g.

```csharp
foreach (var kvp2 in symbolNameToId)
    if (kvp2.Value == currentProcId && kvp2.Key.Contains(".")) { ... }
```

against a 333k-symbol table, once per qualifying call site. The gap is conservative: `after` ran its entire duration alongside `before`, while `before` had ~12 minutes of the machine to itself. Worth calling out in the PR description — it's a real second benefit.

## Static review notes

- Verified the load-bearing assumption: `ClarionParser` records `LineNumber = lineNum` where `lineNum = i + 1` (1-based, the `PROCEDURE` line itself, captured *before* continuation joining), matching the indexer's `i + 1` lookup. Alignment holds including for continuation-spanning definitions.
- `symbolByFile` is still consumed at `CodeGraphIndexer.cs:1309`, so dropping the `currentFileSymbols` local doesn't orphan it.
- Patching both copies is correct — the two `CodeGraphIndexer.cs` files already diverge on `master` (`_resolver` readonly-ness, `libraryPaths` on `Resolve`), so they're genuinely maintained in parallel.

**LGTM.** Only caveat: this is one solution — the right one, but a different codebase could exercise paths it doesn't.

PR #169, opened 2026-08-01
Fix/diagnostics window follow theme + visual improvements + correctness fixes

geircodes:
# fix: diagnostics window follows the CA Editor's theme, plus readability and correctness fixes

## Summary

The LSP status bar and the diagnostics popup window (opened by clicking the error/warning count in the status bar) always rendered in the chat pane's own dark/light setting, which is a separate preference from the CA Editor's own Monaco dark/light setting. If the two didn't match, the diagnostics window would look wrong compared to the editor. On top of that, once matched up, the window itself had a handful of rough edges — some of them plain bugs, some just low-contrast/low-legibility styling — that made it harder to read than it should be.

This PR makes the diagnostics window follow the CA Editor's own theme setting, and fixes those rough edges along the way.

**Scope note:** this started as a purely visual PR, and grew. Working on the window's appearance meant looking at it a lot, and that surfaced three correctness bugs in the same feature — including one where the status bar pill never appeared at all, which made the window unreachable in the most common configuration. Those are fixed here too rather than deferred, since they're the same window and the same code path. They're listed separately below.

## What changed — appearance

- **Theme source**: status bar and diagnostics window now read `CaEditorSettings.MonacoThemeDark` instead of the chat pane's theme. There's no change notification for that setting, so `AssistantChatControl.PollLspUi` (already ticking every 2s) also polls it and re-applies the theme on change — including on the diagnostics window's first-ever open, which had its own separate bug: it was reading the chat pane's theme on construction even after everything else was switched over.
- **Window layout**: filename folded into the window title (`Diagnostics — filename.clw`) instead of its own label row; the text "Refresh" button replaced with a compact icon-only button.

  The filename move is a deliberate space tradeoff, not an accident: the window is 340px tall by default with a 24px toolbar strip, and a separate label row spent roughly 10% of the client area on one short string the title bar shows for free. It costs no information — the window is strictly single-file, and the caption and the rows are populated from the same `filePath` in `PollLspUi`, so the caption can't disagree with the list. (It *could* disagree before this PR, via the refresh button — see the correctness section.)
- **Column header colors**: the header wasn't following the theme at all (plain OS default regardless of dark/light), and the standard `SetWindowTheme` trick for native ListView headers only reliably recolors the background, not the text — that left black-on-near-black in dark mode. Fixed by owner-drawing the header instead, for full control of both.
- **List background**: switched from a low-contrast, slightly-tinted palette to a plainer high-contrast one, and made the list/header background match the window's own background in both themes (previously they were different shades).
- **Grid lines**: native ListView grid lines use a fixed, non-theme-aware system color — shows up as stark white lines in dark mode. Now hand-drawn per-cell using a muted, per-theme color instead.
- **Selection highlight**: was hardcoded to the system blue-highlight/white-text pair, which didn't match the window's palette and overrode each row's severity color (red/amber/blue). Now uses a subtle per-theme background tint and keeps the row's own text color.
- **Hover flicker**: owner-drawing a ListView without enabling double buffering (not exposed publicly on `ListView`, set via reflection) visibly flashes on every repaint, including the ones Windows triggers from mouse-hover tracking alone.
- **Window caption**: added `DWMWA_USE_IMMERSIVE_DARK_MODE` so the native title bar follows dark/light at all, and (Windows 11 22H2+) `DWMWA_CAPTION_COLOR`/`DWMWA_TEXT_COLOR` so it matches the window's own palette exactly rather than the OS's generic dark-mode caption text.
- **Tooltip never showing**: the refresh button's tooltip silently never appeared. `ToolTip` only displays by default when its parent form is the active foreground window, and this non-modal tool window (opened via `Show(owner)` from the docked chat pane) apparently never becomes the active window. `ShowAlways = true` fixes it. (This is very likely also why keyboard input doesn't reach this window — not changed here, just noting it as a related, still-open observation.)
- **Washed-out severity text**: moving cell/header text drawing to manual owner-draw (needed for the header/gridlines/selection work above) introduced its own regression — drawing it with GDI's `TextRenderer.DrawText` rendered the red/amber severity colors visibly paler than their actual value, and passing the real background color through the 7-arg overload (to fix ClearType's background-compositing assumption) didn't fully resolve it either. Switched to GDI+'s `Graphics.DrawString` with a `SolidBrush` instead, which paints the exact requested color via normal alpha blending with no backdrop-composition ambiguity at all.

## What changed — correctness

- **The status bar pill never appeared, so the diagnostics window was unreachable.** `PollLspUi` gated the entire status bar on the *bundled* `LspClient` (`lsp == null || !lsp.IsRunning` → hide). When the IDE's own ClarionLsp addin is the active client, the bundled one is never started at all, so that test failed on every tick and the pill was hidden unconditionally — and clicking the pill is its only trigger. Meanwhile the editor showed squiggles perfectly well, because the squiggle pass and every other consumer already route through `SharedLspBridge`. The two clients also keep *separate* diagnostics caches (`LspClient._diagnostics` vs `SharedLspBridge._sharedDiagCache`), so reading the bundled cache would have come back empty even past the gate. Both the liveness check and the cache read now go through the bridge.

  The bridge has no equivalent of `LspClient.LastActiveFilePath`, so target resolution moved to the active editor. There are two independent Monaco-backed editing surfaces and either can be active: `MonacoClarionEditor` (a Monaco overlay on the native Clarion source editor — the common case, resolved by `EditorService.GetActiveDocumentPath()`), and `ModernEmbeditorViewContent` (the separate CA Editor tab, which exposes neither `FileName` nor `PrimaryFileName`, so `GetActiveDocumentPath()`'s probes always miss it and it needs its own lookup). `ResolveDiagnosticsTarget` tries both, then falls back to `LastActiveFilePath`.

  Preferring the active editor is also a behavioral improvement in its own right: `LastActiveFilePath` follows the last file *any* LSP tool touched, so a chat-driven hover or definition lookup — or a full-solution diagnostics sweep — would silently repoint the pill and the window at a file nobody is editing.
- **The refresh button captioned the window with a stale filename.** `RefreshFromSource` pulled fresh entries from its refresh delegate but reused the cached `_currentFileName` for the caption, so pressing refresh after the target had moved captioned one file over another file's rows. It now takes the file identity from the same source as the entries. (The two reads are safe as a pair: `PollLspUi` sets both together on the UI thread, and the click handler is UI-thread too.)
- **The pill claimed a green "OK" for files it had never heard about.** `SetDiagnostics` had no way to express "don't know yet", so an unanalyzed file was flattened to 0 errors / 0 warnings and rendered with the same green check mark as a genuinely clean file — an affirmative all-clear during exactly the window in which results are still arriving. The distinction was already available and simply being discarded: `GetCachedDiagnostics` returns `null` when nothing has ever been published for a file (the bundled client tracks this as `DiagnosticSet.WasPublished`, whose own comment calls out this exact case; the shared client, by the key being absent), while an authoritatively clean file caches an *empty* list. Added a third, muted pill state for it.

  Deliberately *not* a progress spinner: LSP has no "analysis complete" notification, so nothing can honestly report when a file is finished being analyzed, and a spinner that stopped early would assert a finality it can't know. This only claims what's actually knowable — nothing has been published yet — and self-clears on the next poll tick.
- **Severity text kept the previous theme's palette after a theme switch.** Rows were coloured once, when added, from the `_isDark` in effect at that moment; `ApplyTheme` updated backgrounds, header, gridlines and caption but never the rows already in the list. Switching dark → light therefore left the dark palette's pastel red and amber sitting on the light background, where they read as washed out, and it persisted until the next `UpdateDiagnostics` happened to rebuild the rows.

  This one is **not introduced by this PR** — the per-item colouring has looked like this since the form was written. It was unreachable, because the window's theme used to be fixed at construction, so there was never a live switch to expose it. Making the window follow a theme that changes at runtime is what turns it into a visible defect, so it's fixed here rather than left behind as a parting gift. The colour was recomputable all along; only its input was being discarded. Rows now carry the severity alongside the line number, and both the build path and the recolour path go through one `SeverityColor(severity)` helper so they can't drift apart.
- **The window followed whichever Monaco surface last spoke, not the one being looked at.** `CaEditorSettings.MonacoThemeDark` is a mirror, written whenever any Monaco page posts `themeChanged` — at boot and on each toolbar toggle. Switching tabs posts nothing, so the mirror doesn't move: with two editors open in different themes, the diagnostics window stayed on whichever one was last toggled or last opened, not the active one. Nothing tracked theme per surface to fix this with, so `MonacoEditorControl` gained an `IsDark` property (seeded from its constructor, kept in sync on every `themeChanged` and in `ApplyTheme`), and each editor surface exposes an active-instance accessor for it (`MonacoClarionEditor.ActiveEditorIsDark()`, `ModernEmbeditorViewContent.ActiveViewIsDark()`). `ResolveEditorThemeDark()` tries both, then falls back to the mirror — deliberately the same shape as `ResolveDiagnosticsTarget` above, active surface first, global as fallback.

  The mirror keeps its existing meaning and stays the fallback, so the pre-paint covers/backdrops it was originally built for (dark-mode no-white-flash on embeditor open) are unaffected — this doesn't make theme *shared* across surfaces, only resolves it correctly per consumer. Making it genuinely shared would reverse that original page-authoritative design and change every Monaco window's behaviour, which belongs in its own issue, not here.

## Test plan

- [x] Builds clean
- [x] Deployed and live-tested after each change, in both dark and light mode, including toggling the CA Editor's theme setting while the diagnostics window was open
- [x] Verified the window opens in the correct theme on its very first open, not just after a subsequent toggle
- [x] Confirmed diagnostics content itself (errors/warnings, line numbers, click-to-navigate) unaffected — spot-checked against a file with a real compile error
- [x] Full end-to-end pass on the correctness fixes: open IDE → load solution → pill shows `17W` → click pill → window opens with the 17 warnings → switch to another source file with the window still open → caption follows the new file and the list updates to `1E 7W`, contents correct → close solution → pill disappears → reload the same solution → correct again
- [x] Severity colours stay correct across a live dark ⇄ light switch with the window already populated
- [x] With two editors open in different themes, switching the active tab repaints the pill and window to match within one poll tick (~2s, the existing polling cadence — noticeable but not incorrect; not changed in this PR)

## Related, still open — not addressed here

- The diagnostics window doesn't receive keyboard input (Enter-to-navigate, arrow-key list navigation) even though the handling exists in code. Almost certainly the same root cause as the tooltip bug above — the window never becomes the active foreground window.
- The window is single-file by design. A full-solution sweep reports per-file in chat and doesn't populate this window. Aggregating multiple files into it would be a real feature (File column, per-entry paths, navigate-to-file via the existing `EditorService.NavigateToFileAndLine`) rather than a tweak — nothing is blocked on missing data, since `DiagnosticEntry` has no file field but the cache is keyed by file URI one level up. Deliberately left out of scope.

### DIFF
--- a/ClarionAssistant/AssistantChatControl.cs
+++ b/ClarionAssistant/AssistantChatControl.cs
@@ -958,7 +958,7 @@ namespace ClarionAssistant
                 // Live-follow the Monaco/CA Editor's own theme toggle (independent of this chat
                 // pane's theme) — the page pushes it to CaEditorSettings on every toolbar toggle,
                 // but nothing notifies this control directly, so pick up a change on the next tick.
-                bool monacoDark = CaEditorSettings.MonacoThemeDark;
+                bool monacoDark = ResolveEditorThemeDark();
                 if (_lastMonacoThemeDark != monacoDark)
                 {
                     _lastMonacoThemeDark = monacoDark;
@@ -1076,6 +1076,30 @@ namespace ClarionAssistant
         }
 
         /// <summary>
+        /// Which theme the pill and the diagnostics window should wear: the ACTIVE editor's, not the
+        /// process-wide CaEditorSettings.MonacoThemeDark mirror. That mirror records whichever Monaco
+        /// page most recently booted or toggled — switching tabs posts nothing and moves nothing — so
+        /// polling it alone left the diagnostics window wearing the theme of whatever surface last
+        /// spoke rather than the editor being looked at. Same shape as ResolveDiagnosticsTarget below:
+        /// ask the active editor, fall back to the global. The mirror stays the fallback (and stays a
+        /// mirror), so nothing else that reads it changes behaviour.
+        /// </summary>
+        private bool ResolveEditorThemeDark()
+        {
+            try
+            {
+                bool? overlay = MonacoClarionEditor.ActiveEditorIsDark();
+                if (overlay.HasValue) return overlay.Value;
+
+                bool? modern = ModernEmbeditorViewContent.ActiveViewIsDark();
+                if (modern.HasValue) return modern.Value;
+            }
+            catch { /* fall through to the global mirror */ }
+
+            return CaEditorSettings.MonacoThemeDark;
+        }
+
+        /// <summary>
         /// Which file the pill and the diagnostics window are about. There are TWO independent
         /// Monaco-backed editing surfaces in this addin, and either can be the one the developer is
         /// actually looking at:
@@ -1123,7 +1147,7 @@ namespace ClarionAssistant
                     line => _editorService.GoToLine(line),
                     () => _lastDiagEntries,
                     () => _lastDiagFile);
-                _diagForm.ApplyTheme(CaEditorSettings.MonacoThemeDark);
+                _diagForm.ApplyTheme(ResolveEditorThemeDark());
             }
             _diagForm.UpdateDiagnostics(_lastDiagFile, _lastDiagEntries);
             if (!_diagForm.Visible)
@@ -2059,9 +2083,10 @@ namespace ClarionAssistant
             if (_contentArea != null) _contentArea.BackColor = _isDarkTheme ? Color.FromArgb(12, 12, 12) : Color.White;
 
             // LSP status bar + diagnostics window show diagnostics for the CODE editor, so they
-            // follow the Monaco/CA Editor's own theme (CaEditorSettings.MonacoThemeDark) rather
-            // than this chat pane's own _isDarkTheme — the two are independent settings and can differ.
-            bool monacoDark = CaEditorSettings.MonacoThemeDark;
+            // follow the ACTIVE editor's own theme rather than this chat pane's own _isDarkTheme —
+            // the two are independent settings and can differ. See ResolveEditorThemeDark for why
+            // this asks the active surface instead of reading CaEditorSettings.MonacoThemeDark.
+            bool monacoDark = ResolveEditorThemeDark();
             _lastMonacoThemeDark = monacoDark;
             if (_lspStatusBar != null) _lspStatusBar.ApplyTheme(monacoDark);
             if (_diagForm != null) _diagForm.ApplyTheme(monacoDark);
--- a/ClarionAssistant/AssistantChatControl.cs
+++ b/ClarionAssistant/AssistantChatControl.cs
@@ -971,14 +971,20 @@ namespace ClarionAssistant
                     lsp.OnLspRequest += OnLspRequest;
                 }
 
-                if (lsp == null || !lsp.IsRunning)
+                // Ask the BRIDGE whether an LSP is up, not the bundled LspClient. When the IDE's own
+                // ClarionLsp is the active client the bundled one is never started at all, so
+                // lsp.IsRunning is false and this hid the pill on every tick — even with the CA
+                // Editor showing live squiggles, because the squiggle pass (and every other
+                // consumer) already routes through SharedLspBridge. The two clients also keep
+                // SEPARATE diagnostics caches, so the read below has to go through the bridge too
+                // or it looks in an empty dictionary.
+                if (!SharedLspBridge.IsRunning)
                 {
                     _lspStatusBar.SetDiagnostics(0, 0, hidden: true);
                     return;
                 }
 
-                // Read diagnostics for the last file the LSP operated on
-                string filePath = lsp.LastActiveFilePath;
+                string filePath = ResolveDiagnosticsTarget(lsp);
                 if (string.IsNullOrEmpty(filePath))
                 {
                     _lspStatusBar.SetDiagnostics(0, 0, hidden: true);
@@ -987,7 +993,7 @@ namespace ClarionAssistant
 
                 // Deduplicate diagnostics by (line, message) — the Clarion LSP server
                 // can emit the same diagnostic dozens of times per analysis cycle.
-                var raw = lsp.GetCachedDiagnostics(filePath);
+                var raw = SharedLspBridge.GetCachedDiagnostics(filePath);
                 List<Services.LspClient.DiagnosticEntry> entries = null;
                 int errors = 0, warnings = 0;
                 if (raw != null && raw.Count > 0)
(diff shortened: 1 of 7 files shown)


ClarionLive (commented):
**Approve — one cleanup request before merge.**

I went straight at the change with the most regression potential: swapping the bundled `LspClient` for `SharedLspBridge` in `PollLspUi`. It's safe — the bridge is a strict superset, not a substitution:

`Services/SharedLspBridge.cs:131`
```csharp
get { return Shared != null || (LspClient.Active != null && LspClient.Active.IsRunning); }
```

`Services/SharedLspBridge.cs:707`
```csharp
if (c == null) { var lsp = LspClient.Active; return lsp != null ? lsp.GetCachedDiagnostics(filePath) : null; }
```

Both fall through to the bundled client when no shared client exists, so the bundled-only configuration behaves exactly as before while the shared-client configuration stops hiding the pill unconditionally. The separate-caches point in the description is right, and the delegation means you don't have to reconcile them.

The `null`-vs-empty tri-state is the right fix and the reasoning for *not* shipping a spinner is the best part of this PR — LSP has no "analysis complete" notification, so a spinner that stopped would assert something unknowable. Also good catch that `_lastDiagKnown` has to join the change-detector, or the pill sticks on `○ …` once a file resolves to genuinely clean.

`ResolveDiagnosticsTarget` preferring the active editor over `LastActiveFilePath` is a real improvement independent of the bug — a chat-driven hover silently repointing the pill at a file nobody is editing is worse than the thing it was doing.

**The request: please drop `ClarionAssistant/Version.props` from this PR.**

```
-    <VersionBuild>851</VersionBuild>
+    <VersionBuild>884</VersionBuild>
```

That's the auto-incrementing build stamp picked up from local compiles, not an intentional change. Master's `851` is the stamp from the released v5.5.0 cut, and with six PRs open right now, any two that carry a `Version.props` bump will conflict with each other over a generated file. Everything else here is worth having; this one line just adds merge friction.

Two smaller notes:

- The scope admission is fair, and I wouldn't split it — the three correctness bugs are the same window and the same code path, and the pill-never-appears one made the whole feature unreachable. But this is the one PR of the current six I'd want an explicit live pass on after merge rather than taken on a clean build, simply because so much of it is owner-draw and DWM behaviour that only a screenshot can confirm.
- `DiagnosticsCacheKey`'s doc comment refers to `ModernEmbeditorDiagnostics.Compute`, which #170 renames to `ComputeAsync`. No conflict — just stale text to sweep once both land.

Merge-order suggestion for the current queue: #168 → #171 → #173 → #170 → #172 → this one last, since it's the largest and the `Version.props` line is the most likely thing to collide.

*Reviewed with Claude Code.*

Issue #4, opened 2026-04-09
lib\sqlite-fts5\System.Data.SQLite.dll missing from repo - build fails from fresh clone


msarson:
After cloning the repo and running msbuild, the build fails because \lib\sqlite-fts5\System.Data.SQLite.dll\ is not present in the repository.

The csproj references it directly:
\\\xml
<Reference Include="System.Data.SQLite">
  <HintPath>lib\sqlite-fts5\System.Data.SQLite.dll</HintPath>
</Reference>
\\\

This is a custom FTS5-enabled build of SQLite (not the standard NuGet package), so it can't be restored automatically. It needs to either be committed to the repo under \lib\sqlite-fts5\ or distributed via a custom NuGet feed.

Without it, every developer building from source gets:
\\\
error CS0246: The type or namespace name 'SQLiteConnection' could not be found
\\\

Could you commit the DLL to the repo, or document where to obtain the correct build?

PR #170, opened 2026-08-03
fix(diagnostics): stop the squiggle overlay going blank on slow or premature LSP results

geircodes:
## Summary

The squiggle overlay — and with it every gutter/overview-ruler mark — could render **nothing at all** for a file, even when the diagnostics pill correctly reported an error or warning for that same file moments later.

Four defects on the squiggle diagnostics path, all producing that one symptom. They share a root theme: **the path conflates "no information yet" with "authoritatively zero diagnostics"**, in two separate places, and then compounds it with two latency problems. That's the same conflation #169 fixed for the diagnostics pill (`GetCachedDiagnostics` returning `null` — never published — versus an empty list — published and genuinely clean); the pill survives it anyway because it re-polls every 2 seconds forever, whereas the squiggle path fires once per edit and never looks again.

## 1. A premature empty LSP republish was trusted as final

The Clarion LSP publishes diagnostics progressively after a buffer change: an early batch — sometimes empty — can land before a slower check finishes and republishes the fuller set. The single bounded `SharedLspBridge.WaitForDiagnostics(lspFileName, 1500, true)` call in `ModernEmbeditorDiagnostics` could catch that early batch and treat it as the final answer.

Reproduced directly against a running LSP session: a diagnostics request raced an authoritative-looking `{"pending":false,"count":0}` result **twice in a row** before the real, complete set — including a warning that needs cross-file scope resolution to compute — arrived on a third query. Same file, same content, no edit in between.

`WasPublished` cannot distinguish this case, because a premature batch *is* published (`WasPublished=true`, `Entries=[]`) — identical to a genuinely clean file by that signal alone.

**Fix:** only an *empty* result gets extra scrutiny — a non-empty one already found something real and returns immediately, so files with existing diagnostics see no added latency. For an empty result, re-read the diagnostics cache over a short settle window and take whatever lands. The re-read is passive: it inspects the cache the publish handler populates, and never issues another request.

## 2. A timed-out round-trip actively wiped the marker set

`requestFromHost` resolves `null` when the host doesn't reply in time, and `refreshDiagnostics` folded that straight into an empty marker list:

```js
var markers = (resp && resp.markers) || [];
```

Since `setModelMarkers` **replaces** the entire set, a slow round-trip didn't merely fail to add a squiggle — it erased every existing one, plus its gutter and overview-ruler marks, with no retry until the next edit. That is the worst available reading of "don't know."

**Fix:** a `null` response now leaves the rendered markers alone; Monaco already tracks them across subsequent edits, and the next successful round-trip replaces them wholesale. An authoritative empty list still clears, as it should.

## 3. The client's timeout sat below the host's worst case

That timeout was a fixed 4000ms, shared with the interactive completion and hover callers — but below the diagnostics path's own worst case (the settle window described above, plus the buffer sync and marshalling a whole-file buffer across the WebView2 boundary in each direction). So the client could discard a slow but **successful** analysis.

**Fix:** `requestFromHost` takes a per-call timeout, defaulting to the previous 4000ms so completion and hover keep their short interactive budget. Only the diagnostics call opts into a longer one. Raising the shared constant instead would have made an unresponsive LSP hang hover and completion too — a regression in the opposite direction.

## 4. The settle loop blocked thread-pool threads

That same settle window used `Thread.Sleep`, and both call sites dispatch `Compute` through `Task.Run`. Keystrokes spaced further apart than the page's own 600ms debounce each dispatch their own request, so several run concurrently — each parking a pool thread for the whole settle window. Past the pool's core-count baseline .NET injects replacements at only ~1–2/sec, so queuing delay compounds precisely when requests overlap, which is the same slow-machine case the settle window exists to serve.

**Fix:** `Compute` becomes `ComputeAsync` and awaits `Task.Delay`. Both call sites already ran off the UI thread, so nothing moves on or off it; `ConfigureAwait(false)` keeps the continuation off the captured context exactly as before.

## Verification

- Clean build, no new warnings (`ClarionAssistant.csproj`, VS2022 MSBuild toolset), with the branch built standalone off `master`.
- Live-tested for the premature-republish case: an identifier declared elsewhere in the workspace but out of scope at the point of use — so ruling it out requires the slower cross-file check — previously never squiggled, while a similar identifier appearing nowhere in the workspace always did. After the fix both squiggle correctly, gutter marks included, and the previously-slow transition is immediate. No added latency observed on files that already carry diagnostics.
- The remaining three only manifest under timing conditions. Exercised live by typing at roughly one character per second — slower than the editor's own 600ms debounce, so every keystroke dispatches its own request and several overlap: markers appeared consistently, the gutter stayed stable, and there was no flashing or marker loss. The specific worst-case triggers (a round-trip exceeding the client timeout; thread-pool starvation under sustained overlap) are not reproducible on demand and were verified by code reading.

## Known remaining gap, deliberately not in this PR

`refreshDiagnostics` has no request sequencing: each debounce cycle dispatches its own request, and whichever *response* arrives last wins, regardless of dispatch order. Because the diagnostics cache is keyed per file with no version, an older request can return *newer* LSP data than the buffer it was computed against, so markers can land a few lines off until the next edit corrects them.

That is a distinct bug — different root cause, different symptom, and independent of everything above — so it belongs in its own change. Worth noting for whoever picks it up: `Terminal/monaco-embeditor.html` already implements exactly the needed guard for saves (`if (msg.savedSeq == null || msg.savedSeq === editSeq) doCancel();`), so the pattern is established there rather than novel.

### DIFF
--- a/ClarionAssistant/MonacoClarionSourceEditor.cs
+++ b/ClarionAssistant/MonacoClarionSourceEditor.cs
@@ -984,10 +984,14 @@ namespace ClarionAssistant
         {
             int reqId; string buffer; List<int[]> ranges;
             if (!ParseDiagRequest(rawJson, out reqId, out buffer, out ranges)) return;
-            System.Threading.Tasks.Task.Run(() =>
+            System.Threading.Tasks.Task.Run(async () =>
             {
                 var markers = new List<Dictionary<string, object>>();
-                try { markers = ModernEmbeditorDiagnostics.Compute(_filePath, buffer ?? "", ranges, null, embedSlotChecks: true); }
+                try
+                {
+                    markers = await ModernEmbeditorDiagnostics.ComputeAsync(
+                        _filePath, buffer ?? "", ranges, null, embedSlotChecks: true).ConfigureAwait(false);
+                }
                 catch (Exception ex) { MonacoSpikeLog.Write("overlay diagnostics error: " + ex.Message); }
                 editor.PostResponse(reqId, new Dictionary<string, object> { { "markers", markers } });
             });
--- a/ClarionAssistant/Services/ModernEmbeditorDiagnostics.cs
+++ b/ClarionAssistant/Services/ModernEmbeditorDiagnostics.cs
@@ -162,11 +162,7 @@ namespace ClarionAssistant.Services
                     int off = (lspContext != null) ? lspContext.LineOffset : 0;
                     SharedLspBridge.EnsureBufferSynced(lspFileName,
                         (lspContext != null) ? lspContext.WrapBuffer(buffer) : buffer);
-                    var wait = SharedLspBridge.WaitForDiagnostics(lspFileName, 1500, true);
-                    List<LspClient.DiagnosticEntry> entries =
-                        (wait != null && !wait.Pending && wait.Entries != null)
-                            ? wait.Entries
-                            : (SharedLspBridge.GetCachedDiagnostics(lspFileName) ?? new List<LspClient.DiagnosticEntry>());
+                    List<LspClient.DiagnosticEntry> entries = WaitForSettledDiagnostics(lspFileName);
 
                     foreach (var d in entries)
                     {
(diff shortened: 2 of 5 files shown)


ClarionLive (commented):
**Approve — merge as-is.**

Four genuinely distinct defects that happen to share a symptom, each argued from its own mechanism rather than lumped together. The framing — "the path conflates *no information yet* with *authoritatively zero*, then compounds it with two latency problems" — is the right diagnosis, and #2 in particular (a timeout folding into `markers = []` when `setModelMarkers` replaces the whole set) is the kind of bug that stays invisible precisely because its symptom looks like success.

Checked the mechanical risk of the async conversion: `ModernEmbeditorDiagnostics.Compute` had exactly two callers repo-wide —

- `MonacoClarionSourceEditor.cs:990`
- `Terminal/ModernEmbeditorViewContent.cs:2959`

— and both are updated here. No orphaned call sites, and both were already inside `Task.Run`, so nothing moves on or off the UI thread.

Two consequences to be aware of, neither a reason to hold this:

1. A genuinely clean file now pays the full settle window on every refresh — 1500ms initial wait plus up to 5 × 400ms — before returning its empty list, because the loop only exits early on a *non-empty* batch. Invisible in practice (there's nothing to draw either way), and `Task.Delay` keeps the overlapping requests cheap, which is exactly the point of fix #4. Worth knowing it's the common case that pays, though, not the rare one.
2. With the timeout no longer clearing, a file that genuinely goes clean while the host is timing out keeps its stale squiggles until the next successful round-trip. That's the right trade against wiping real markers, and the code says so — just noting it's a real behaviour change, not a pure win.

The "known remaining gap" section is doing real work — please open an issue for the request-sequencing/marker-drift problem so it doesn't only exist in a merged PR description. The pointer to `monaco-embeditor.html`'s existing `savedSeq` guard as the established pattern is exactly what whoever picks it up will want.

Merge-order notes, since there are five other PRs open against this file set:
- #173 also edits `MonacoClarionSourceEditor.cs` in overlapping context — this PR's hunk ends at line 993, #173's starts at 993 (the `OnClipboard` stub immediately after `OnDiagnostics`). Whichever lands second may need a rebase.
- #169 adds a comment referring to `ModernEmbeditorDiagnostics.Compute`, which this PR renames to `ComputeAsync`. No merge conflict (different files), just stale doc text to sweep once both land.

*Reviewed with Claude Code.*

PR #172, opened 2026-08-03
fix(completion): bare-prefix CodeGraph merge leaks procedure-local variables solution-wide

geircodes:
## Summary

Follow-up to the `?Ctrl` field-equate bare-prefix merge guard (PR #159 — see
[pr-description-field-equate-bare-prefix-merge-guard.md](pr-description-field-equate-bare-prefix-merge-guard.md)).
That PR guarded `MergeBarePrefixCompletions()`'s own qualified-context check, but a *different*
merge step inside the same function — `MergeDbBarePrefix()`, which backfills bare-word completion
from the solution-wide CodeGraph database — had no equivalent guard against surfacing
procedure-private symbols.

Found while investigating a Clarion-Extension bug report (missing `INCLUDE` keyword completion;
see companion Clarion-Extension PRs #393/#394): even after both upstream LSP fixes deployed,
completion at the top of a PROGRAM file still surfaced variables like `IncludeAddress` — a real
variable, but declared as a procedure-local inside an entirely unrelated procedure, in an entirely
different file.

## Root cause

`MergeDbBarePrefix()` calls `CodeGraphProvider.FindSymbols()`, which does a plain substring `LIKE`
match across the whole solution's indexed `.codegraph.db` — no file or scope awareness at all.
`ClarionParser` already tags every indexed symbol with a `Scope`: `"local"` (procedure-private),
`"module"` (file-scope), `"global"`, `"class"`, `"virtual"` — but `MergeDbBarePrefix()` never reads
that field. Every `"local"` symbol matching the typed prefix gets merged into completion exactly
like a genuine global, regardless of which procedure or file the cursor is actually in.

## Fix

One-line guard added to the merge loop, `Services/SharedLspBridge.cs` (`MergeDbBarePrefix`):

```csharp
if (string.Equals(s.Scope, "local", StringComparison.OrdinalIgnoreCase)) continue;
```

Placed before the existing prefix/dedup checks. The sibling call for the ClarionGraph static
library DB (`bareNamesOnly: true`) is unaffected — those symbols are never scoped `"local"` in the
first place, so the guard is a no-op there.

## Testing

Clean `Debug-C` build (MSBuild, `ClarionRoot=D:\_dev\C111`), only `Services/SharedLspBridge.cs`
touched. No automated test harness exists for this addin (same as PR #159) — verification is
build-clean + live-verify in the IDE.

### DIFF
--- a/ClarionAssistant/Services/SharedLspBridge.cs
+++ b/ClarionAssistant/Services/SharedLspBridge.cs
@@ -1899,6 +1899,13 @@ namespace ClarionAssistant.Services
                     foreach (var s in syms)
                     {
                         if (s == null || string.IsNullOrEmpty(s.Name)) continue;
+                        // A "local" symbol is procedure-private (CodeGraph's own scoping, set by
+                        // ClarionParser) — never a valid completion candidate outside the procedure
+                        // that declared it, let alone from a different file across the whole
+                        // solution. FindSymbols() has no scope awareness (plain name LIKE match), so
+                        // this merge must filter it out itself instead of surfacing every same-prefix
+                        // local from every procedure in every file as if it were global.
+                        if (string.Equals(s.Scope, "local", StringComparison.OrdinalIgnoreCase)) continue;
                         if (!s.Name.StartsWith(prefix, StringComparison.OrdinalIgnoreCase)) continue; // true prefix only
                         if (bareNamesOnly && s.Name.IndexOf('.') >= 0) continue; // ClassName.Method → member-access only
                         if (!seen.Add(s.Name)) continue;


ClarionLive (commented):
**Approve in principle — one question before merging.**

The guard is correct, and I specifically checked the thing most likely to break: whether filtering `Scope == "local"` out of the DB merge costs you completion of the locals in the procedure you're actually standing in. It doesn't. Those come from step (1) of `MergeBarePrefixCompletions`:

```csharp
// (1) Local variables (depth-aware) from the enclosing routine + procedure DATA. Never in the
// CodeGraph (not cross-project) + must reflect unsaved edits, so parse the live buffer. Phase 2.
if (lines != null) MergeLocalVarCompletions(primary, seen, prefix, lines, line);
```

So the buffer parse already owns in-scope locals by design, and `MergeDbBarePrefix` was only ever adding *other procedures'* locals on top. Filtering them is exactly right. `CgCompletionDetail`'s `isLocal ? "local" : "global"` branch was effectively documenting the bug — it knew locals were arriving and just labelled them.

**The question:** `ClarionParser` also tags ROUTINEs as `Scope = "local"`, not just variables —

`CodeGraph/Parsing/ClarionParser.cs:505-514`
```csharp
result.Symbols.Add(new ClarionSymbol
{
    Name = routineName,
    Type = "routine",
    ...
    Scope = "local"
});
```

— and I couldn't find any C#-side merge that offers routine *names* for completion: there's `MergeLocalVarCompletions`, `MergeModuleVarCompletions` and `MergeLocalProcedureCompletions` (MAP prototypes + in-buffer procedure implementations), but nothing analogous for routines. Which suggests `DO MyRou…` was previously being completed via this very DB path — correctly for the current file, alongside every routine in every other file, which is the bug you're fixing.

So: does the LSP server itself supply routine names in completion? If it does, this is a non-issue and I'd merge as-is. If it doesn't, the guard may want to be same-file-aware for `Type == "routine"` rather than blanket — or routines want their own buffer-parsed merge alongside the other three, which would be the more consistent shape given locals are already handled that way.

Happy to be told the LSP has this covered — I only have the C# side in front of me.

*Reviewed with Claude Code.*

ClarionLive:
Closing the loop on my routine question above — no answer needed, it's resolved.

**Your fix was right and stays exactly as merged.** The guard is correct, and the reasoning for it holds: procedure-private symbols have no business being offered solution-wide, and in-procedure locals were never at risk because they come from the live-buffer parse at step (1), not the DB.

The routine concern turned out to be real, though. Confirmed in live testing right after merge: `DO Refr` stopped completing `RefreshWindow`. `ClarionParser` scopes ROUTINEs `"local"` right alongside procedure-local variables (`CodeGraph/Parsing/ClarionParser.cs:514`), so they went out with the leak your guard was closing.

The interesting part is what the *old* behaviour actually was. Typing `DO ref` returned methods from an unrelated `Reflection` class — and **not** the `RefreshWindow` routine three lines up. So the DB was never really answering that context correctly; it was answering it with every same-prefix routine in the solution plus a pile of globals, and the one routine that was actually in scope was competing with all of it. Your guard didn't break `DO` so much as reveal that it had never worked properly.

Fixed in `9b87b31` on master, and it follows the precedent your PR's own reasoning established:

- `MergeRoutineCompletions` reads routine labels **from the live buffer**, scoped to the enclosing procedure — which is a routine's real visibility in Clarion. Same argument that already puts local *variables* on the buffer-parse path rather than in the DB merge, and it means an unsaved routine completes too, which the DB could never do.
- `DO <prefix>` became its own completion context that returns routines *only* and merges no global source at all. `DO` takes a routine label and nothing else, so `Reflection`'s methods can't come back.
- Scoping uses the same never-blank-a-working-list guard as the existing member-access and colon-qualifier passes — a prefix with no routine behind it keeps whatever the LSP offered rather than showing an empty popup.

Verified live: `DO ref` now offers `RefreshWindow` and nothing from `Reflection`.

Net effect is that your PR fixed the leak *and* the DO path came out better than it was before it — worth knowing if the same `Scope` field turns up in other merges.

*Follow-up written with Claude Code.*

Issue #140, opened 2026-07-22
Monaco editor makes the IDE output-pane unavailable


geircodes:
When a monaco editor is the active open document, it denies the output pane (Build output/Generator output/Debug) focus, The editor steals keypresses and mouse clicks.

When I open the output pane I can't select text, not even with the rightclick-select all. CTRL-A selects all the text in the editor window. 
The only events that this pane gets are mouse scroll, so I can use the scroll wheel to switch between build/generator/debug, and I can scroll in the text.

It is in the annoyance category as I some times want to look in the build output.
I had Claude look at it but it was unsuccessful, ended up with focus problems related to WebView2.

The workaround is to open a file in the clarion editor and have that as the active window, then I can access the Output as I am used to.

I checked other panes too: 
Search Results pane works fine, If I have searched in a clarion editor and pressed "find all", the search results list is populated, and mouseclicks works for navigating to the result
Clipboard Ring works fine.

ClarionLive:
Confirmed and ticketed. This is the WebView2 focus family — the embedded browser that hosts the Monaco editor is winning the input fight when you click into the Output pane (which is also why scroll still works: wheel events route by hover, not focus). Interesting data point that Search Results behaves — that asymmetry should help us find the right hook. The likely fix is detecting activation of a non-editor pad and explicitly releasing WebView2 focus to it; we'll be careful not to regress the editor-focus-on-tab-switch work while at it. Your workaround (make a native editor tab active first) is the right one in the meantime.

peterparker57:
Fix landed on master (39fb149) and verified in-IDE — the Output pane is fully interactive again (click, drag-select, Ctrl+A, Build/Generator/Debug switching) while a CA Editor tab is the active document, with no regression to the editor-focus-on-tab-switch behavior.

Root cause, confirmed with live tracing: it was friendly fire from our own tab-activation focus fix. The IDE never hands a WebView2-hosted view keyboard focus on a tab switch, so CA hooks the workbench's WindowSelected event and claims focus for the editor (with a verify-and-retry ladder). It turns out the IDE re-fires that event for every click into a docked pad while a Monaco document stays active — so each click into the Output pane triggered a fresh focus claim that yanked the keyboard straight back into the editor. Your observation that scrolling still worked was the perfect clue: wheel events route by hover, everything else routes by focus. (Search Results behaved because interacting with it doesn't re-fire the event the same way.)

[...]

PR #173, opened 2026-08-04
Fix/ca editor CTRL-X did not copy to clipboard 

geircodes:
## Summary

`MonacoClarionEditor` (the CA Editor's Monaco/WebView2 overlay, `MonacoClarionSourceEditor.cs`) implements `IMonacoEditorHost` with two callbacks that were left as empty stubs from the original overlay-convergence spike and never wired up:

- **`OnClipboard`** — `doClarionCut`'s Ctrl+X (`monaco-embeditor.html`) always posts `{action:"clipboard"}` with the cut/removed text before deleting it from the buffer, for both a selection cut and a no-selection whole-line cut. The CA Editor never forwarded that message to `Clipboard.SetText`, so **Ctrl+X deleted the text without ever putting anything on the Windows clipboard.**
- **`OnFocusEditor`** — `installDropInsert`'s native OS drop handler posts `{action:"focusEditor"}` after a Data-pad field drag-drop lands directly on the WebView2 surface, since OS keyboard focus stays on the Data pad (a separate WebView2) after the drop. Left unhandled, the CA Editor's own workbench tab never reclaimed activation after such a drop.

## Root cause

Both are the same bug class: the JS side of the shared `monaco-embeditor.html` protocol posts these messages unconditionally (not gated to embed mode), but this particular host (`MonacoClarionEditor`) never implemented its half of the contract. The CA Embeditor's own host (`ModernEmbeditorViewContent`) already implements both correctly — `HandleClipboard` calls `Clipboard.SetText`, and `OnFocusEditor` calls `BringToFront()`.

## Fix

- `OnClipboard`: ported `ModernEmbeditorViewContent.HandleClipboard`'s logic — deserialize the `text` field and call `Clipboard.SetText`, with the same empty-string guard (`Clipboard.SetText` throws on `""`, so an empty cut is written as `" "`). Also logs the written char count (and a distinct message if a payload ever arrives with no `text` field) — added while chasing a report that looked at first like a selection-specific clipboard bug, see Verification below.
- `OnFocusEditor`: activates this editor's own workbench tab via the same `SelectWindow` reflection idiom already used elsewhere in this file (`OnWorkbenchWindowSelected`, `CloseWorkbenchTab`) — `MonacoClarionEditor` doesn't have a `BringToFront()`-style helper like the embeditor's view content, since it's a `ClarionEditor` subclass rather than an `AbstractViewContent`, but it already caches `_wbWindow` for the same reflection call elsewhere.

## Two stubs deliberately left alone

The same interface implementation also has empty `OnCancel` and `OnOpenSource` stubs. Traced both in `monaco-embeditor.html`: the Cancel button and the clickable embed header are only ever shown/wired in embed-overlay mode (`has-embheader`) — file mode (which the CA Editor always runs in) explicitly keeps closing via the tab instead and never renders that header. Neither message is ever posted from the CA Editor, so these two are genuinely unreachable dead code, not a second instance of this bug — left untouched rather than "fixed" for something that can't happen.

## Verification

- Clean build via VS2022's MSBuild (`ClarionAssistant.csproj`, Debug, ClarionVersion=11).
- **Live-tested in the running Clarion 11.1 IDE** (deploy.ps1 rebuild+redeploy). Confirmed via `monaco-spike.log`:
  - Ctrl+X with no selection (whole-line remove): clipboard populated correctly.
  - Ctrl+X with an active selection: clipboard populated correctly (log shows `OnClipboard: wrote 47 chars to clipboard`, matching the selection cut, and a later `wrote 7 chars` for a follow-up line cut) — an initial round of this same test appeared to fail, but that was the git clone having been left checked out on an unrelated in-progress branch that never had this fix merged in, not a real code issue; redeploying from the actual fix branch resolved it.

### DIFF
--- a/ClarionAssistant/MonacoClarionSourceEditor.cs
+++ b/ClarionAssistant/MonacoClarionSourceEditor.cs
@@ -993,8 +993,23 @@ namespace ClarionAssistant
             });
         }
 
-        // The rest stay inert for now (Step 6 wires Ctrl+D through the embeditor's openDesigner path).
-        void IMonacoEditorHost.OnClipboard(MonacoEditorControl editor, string rawJson) { }
+        /// <summary>{action:"clipboard"} — Clarion-style Ctrl+X (doClarionCut in monaco-embeditor.html) posts the
+        /// cut/removed text here before deleting it from the buffer. This host used to leave the message unhandled,
+        /// so Ctrl+X in the CA Editor silently deleted the line without ever reaching the Windows clipboard —
+        /// unlike the embeditor's own OnClipboard (ModernEmbeditorViewContent.HandleClipboard), which does the
+        /// same Clipboard.SetText call. Clipboard.SetText throws on an empty string, hence the " " fallback,
+        /// matching the embeditor's handling exactly.</summary>
+        void IMonacoEditorHost.OnClipboard(MonacoEditorControl editor, string rawJson)
+        {
+            try
+            {
+                var ser = new JavaScriptSerializer { MaxJsonLength = int.MaxValue };
+                var data = ser.DeserializeObject(rawJson) as Dictionary<string, object>;
+                string text = (data != null && data.ContainsKey("text")) ? (data["text"]?.ToString() ?? "") : null;
+                if (text != null) Clipboard.SetText(text.Length == 0 ? " " : text);
+            }
+            catch (Exception ex) { MonacoSpikeLog.Write("OnClipboard error: " + ex.Message); }
+        }
 
         // CA Find pad protocol (GitHub #66) — the broker routes to/from the dockable pad.
         void IMonacoEditorHost.OnCaFind(MonacoEditorControl editor, string action, string rawJson)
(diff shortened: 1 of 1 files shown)


ClarionLive (commented):
**Approve — merge as-is.**

Straightforward "host never implemented its half of the contract" fix, and the port stays faithful to the working implementation rather than reinventing it — `OnClipboard` matches `ModernEmbeditorViewContent.HandleClipboard:3476-3484` line for line, including the `" "` fallback for the empty cut (`Clipboard.SetText` throws on `""`).

Checked the things that could bite:

- Both required usings are already in `MonacoClarionSourceEditor.cs` — `System.Web.Script.Serialization` (line 7) for `JavaScriptSerializer` and `System.Windows.Forms` (line 8) for `Clipboard`.
- `Clipboard.SetText` needs an STA thread; `WebMessageReceived` fires on the UI thread, so this is on the same footing as the embeditor's handler that's been working in production.

The "two stubs deliberately left alone" section is the right call and the right level of evidence — tracing `OnCancel`/`OnOpenSource` to `has-embheader` and confirming file mode never renders it is a much better answer than implementing something unreachable for symmetry.

One thing worth watching when this is live, not a code objection: `OnFocusEditor` calls `SelectWindow`, which activates the workbench tab. Tab activation in this fork re-fires `WindowSelected` per pad click, which is what `EditorFocusGuard.FocusInForeignPad` exists to gate (GH #140). A drop-triggered `SelectWindow` should be a one-shot and I'd expect no interaction, but the Data-pad drag-drop path is precisely where a focus ping-pong would show up. Worth one explicit drag-drop-into-CA-Editor pass after merge to confirm focus lands and stays.

*Reviewed with Claude Code.*

geircodes:
Post-merge focus test done, as suggested — focus lands and stays.

**Test:** dragged a text selection from the CA Find pad's search entry field onto the open Monaco editor. Text inserted at the drop point, editor took focus, typing continued in the editor rather than the search field. Same result from the COM Control Browser pad's info-text. Ctrl+X re-verified first as a build-identity canary (`OnClipboard: wrote 46 chars to clipboard`) to confirm the deployed build contained the fix.

**On the `SelectWindow` / GH #140 interaction — no ping-pong.** The guard engages and declines once per interaction:

```
15:19:49.791  focus-guard: stand down — focus in ...PadContentWrapper
15:19:49.791  select-hook fired: STAND DOWN (focus in foreign pad) (<file>)
15:20:41.061  focus-guard: stand down — ...
15:20:42.948  focus-guard: stand down — ...
15:21:01.170  focus-guard: stand down — ...
```

Pairs are seconds apart — discrete interactions, not a loop. (The fork's known doubled fire looks different: two lines 13 ms apart, as documented at `MonacoClarionSourceEditor.cs:2093`.) `SelectWindow()` activates the tab, the re-fired `WindowSelected` hits the guard, the guard correctly declines the redundant `_editor.FocusEditor()`, and the JS-side `ed.focus()` has already placed the caret. Your "one-shot, expect no interaction" read was right.

**Correction to the PR description:** it says the fix covers a Data-pad *field* drag-drop. That's inaccurate. `modern-data-pad.html:1455` (`row.draggable = !dragMeta;`) selects between two mutually exclusive drag mechanisms:

- Dictionary **column** (`:2001`) and **variable** (`:2321`) rows pass `dragMeta`, so `draggable` is off and the drag goes through the host-owned `FieldDropService` mouse hook — no DOM `drop`, never runs `installDropInsert`, never posts `{action:'focusEditor'}`.
- Routine (`:2116`), procedure (`:2164`) and generic insert rows (`:1610`) use the HTML5 OLE drag, and those are the ones reaching `OnFocusEditor`.

So an actual field drag never exercised this code. Real scope is procedure/routine/generic drags plus any in-IDE `text/plain` drag source.

Out of scope but noted: a cross-application drag inserts the text but leaves keyboard focus in the source app — expected, since `ed.focus()`/`SelectWindow()` only move focus within the IDE's own window. Also a pre-existing `FocusEditor: focus never landed after retries` immediately after `initial-open focus: active=True` on file open, unrelated to this path.

Issue #179, opened 2026-08-05
CA Embeditor close: native Access Violation in Clarion's Generator ("Internal error: Restore Procedure")


asantarelli:
## Summary

Closing the CA Embeditor overlay can crash `Clarion.exe` outright with a **native Access Violation inside Clarion's own Generator**, not a .NET exception. Confirmed via Clarion's own internal exception dialog (transcribed below, not a screenshot), not a Windows WER/Application-log event — the Windows Application log has nothing for the crash moment; System log is unrelated background noise (RsFx0800 driver events).

## Crash detail

```
Exception occurred at address 13FAC9B5
Exception code C0000005: Access Violation
Process PID 7016  Image: D:\Clarion11\bin\Clarion.exe  Thread 3  Handle=00000F0C  TID 19428

Exception parameters:
00000000
00000C33

EAX=00000C33 EBX=000007D0 ECX=1408E3C4 EDX=1408176C
ESI=1D5768B0 EDI=1E3CEDD7 EBP=1E3CF3BC ESP=1E3CED74

Call Stack:
13FAC9B5
13FDF22B
0A93DD39
777EE12B
777EE0B1
```

A second dialog appears alongside it, titled **"Generator"**, reading:

> Internal error: Restore Procedure

No symbols available for the call-stack addresses (no matching PDBs on our end) — flagging in case anyone on the Clarion/SoftVelocity side can resolve them against a Generator build.

## Repro

1. Open a procedure's embed via the CA Embeditor (overlay mode — right-click a procedure → Embeditor Source).
2. Edit code in an embed slot.
3. Close/save the embed (native Save & Exit gesture, or the CA Embeditor's own save-and-close).
4. Intermittent — not every close reproduces it, but it recurs reliably enough over a session (several times in one afternoon of normal editing) to be a serious stability concern, not a one-off.

## Why this looks Generator-side, not WebView2/CA-side

We spent a full session chasing this as a suspected WebView2-teardown race in `MonacoEditorControl.PostJson`/`PostString` (a background diagnostics continuation posting into a WebView2 control mid-`Dispose()` — real bug, fixed separately, see the companion PR). That fix did NOT stop this crash from recurring. The dialog itself settles it: this is inside Clarion's own **Generator** component doing something called **"Restore Procedure"**, at the moment the embed closes — native code we don't call directly by name anywhere in ClarionAssistant.

## Likely connection to #142

[...]

Issue #142, opened 2026-07-22
Clarion became unstable, and Embeditor Took Over


BoxSoft:
I've currently got both Clarion Assistant Monaco options disabled. I was in the regular embeditor, editing code for a while. Suddenly the screen changed to the Monaco embeditor, then a few seconds later Clarion crashed. I restarted Clarion and confirmed that the CA options have the Monaco options disabled.

BTW, I installed 5.4, but the CA Assistant panel says it's still 5.2.691.

In general, I've noticed that Clarion is far more fragile than it was before. Even if I don't go into the Screen editor and visit embeds from there, Clarion crashes at least once per hour. I suspect all of this baggage has made it more unstable than it was before. 😟

ClarionLive:
Thanks Mike — the version line is the key clue here: **the panel reads its version from the loaded DLL, so if it says 5.2.691, Clarion is genuinely still running the 5.2.691 addin**, not 5.4. That would explain both symptoms:

- The "Monaco options disabled but the embeditor took over anyway" behavior matches how 5.2 triggered the overlay (it hooked Clarion's own Embeditor Source open). The option gating and the embeditor open/close lifecycle were substantially reworked in 5.3/5.4, along with several crash fixes in exactly that area.
- The general instability you're seeing is being reported against a build that's two releases behind.

We verified the 5.4 installer shipped fresh 5.4.787 binaries for all four Clarion editions (C10/C11/C11.1/C12), so the install itself was good — it just didn't land in the Clarion you're running. The two usual causes:

1. **The DLL was locked during install.** If any Clarion.exe was still alive (even a zombie that closed its window but never exited — check Task Manager → Details for Clarion.exe), the installer can't replace the addin and your old files stay in place. We've seen third-party apps (screen/audio tools) hold handles to a dead Clarion process and keep the DLL locked until reboot.
2. **The installer targeted a different Clarion root** than the one you launch (if you have more than one install, or a non-standard path).

Could you check:

1. Which Clarion version/path you're running, and the file version of `<ClarionRoot>\accessory\addins\ClarionAssistant\ClarionAssistant.dll` (right-click → Properties → Details). I'd expect it to say 5.2.691.0.
2. Then: close Clarion, confirm no Clarion.exe remains in Task Manager (reboot if one won't die), re-run the 5.4 setup, and confirm the panel shows **5.4.787** on next launch.

[...]

Issue #154, opened 2026-07-27
Ctrl+J snippets: split a selection into multiple parameters


asantarelli:
The Code Snippets (Ctrl+J) `${SELECTED}` token substitutes the WHOLE selection as one blob, with no way to pull a structured selection like `Clientes,CLI:CLI01` apart into separate insertion points — e.g. to build `Access:Clientes.Fetch(CLI:CLI01)` from a `GET(Clientes,CLI:CLI01)` call, so an AHK-style macro user could get the same result from a plain Ctrl+J snippet instead.

`parseSnippetBody`'s tab-stop grammar (`monaco-embeditor.html`) is a deliberately minimal custom parser (`$N` / `${N:default}`), not Monaco's native snippet engine, so there's no regex/variable-transform escape hatch to reach for either.

A PR follows: `${SELECTED:N}` (1-based) substitutes the Nth comma-separated part of the selection, trimmed.

peterparker57:
Shipped — merged via #155 (thanks @asantarelli).

`${SELECTED:N}` gives you the Nth comma-separated part of the current selection (1-based, trimmed), alongside the existing `${SELECTED}` for the whole selection. Mix them freely in one body:

```
Access:${SELECTED:1}.Fetch(${SELECTED:2})
```

Select `Clientes,CLI:CLI01` (say, from inside a `GET(Clientes,CLI:CLI01)` call), press **Ctrl+J**, and that expands to:

```
Access:Clientes.Fetch(CLI:CLI01)
```

Details worth knowing:

- **Out-of-range indices** substitute empty rather than erroring.
- **With nothing selected**, each *distinct* `${SELECTED:N}` spelling becomes its own fillable tab-stop instead of a blank hole — and repeats of the same spelling share one index and mirror as you type, exactly like repeated `$N` stops.
- **Wrap-style snippets stay out of trigger-word completion.** Typing a trigger has already destroyed any selection by the time you finish, so bodies referencing `${SELECTED}` or `${SELECTED:N}` remain Ctrl+J-only — that filter now recognises both spellings.

There is a worked example in the snippet editor: click the **?** next to the Body field.

Landed on `master` and verified in Clarion 12. It will appear in the next numbered release; the README's **What's New (Unreleased)** section covers it in the meantime.

asantarelli:
Thanks!

ClarionLive:
🚀 **Shipped in v5.5.0.**

[...]

PR #180, opened 2026-08-05
fix(embeditor): WebView2 post-after-dispose race on closing a CA Embeditor tab

asantarelli:
## Summary

- Fixes a real, distinct race in `MonacoEditorControl.PostJson`/`PostString` (the single chokepoint every host→page message — diagnostics, hover, completion, save results, designer messages, ... — already funnels through): a background continuation can still be posting a response after `DetachOverlay()`/`Dispose()` has started tearing the WebView2 control down.
  - `_webView != null` doesn't detect disposal — `Dispose()` never nulls the field, only `IsDisposed` flips.
  - `InvokeRequired` on a control whose window handle was just destroyed returns `false` — not because the call is already on the UI thread, but because it has no handle left to compare against — so the post fell through to calling the WebView2/CoreWebView2 COM object (STA) directly from a thread-pool thread instead of marshalling.
  - Fix: check `IsDisposed` before trusting `InvokeRequired` at all, both at the outer guard and again inside the marshalled action.
- v5.6's diagnostics settle-loop (#170) widened the window for this considerably — a diagnostics round-trip that used to resolve within a bounded 1.5s wait can now take several seconds across multiple thread-pool hops, so it's far more likely to still be in flight when a tab closes.

## Scope — please read before assuming this closes out embed-close crashes

This is **not** a full fix for embed-close instability. While chasing a recurring crash-on-close report, this fix stopped one clean reproduction, but the same symptom reproduced again afterward — this time caught as Clarion's own internal exception dialog, a **native Access Violation inside the Generator** ("Internal error: Restore Procedure"), not a .NET/WebView2 issue at all. Filed separately as #179, with the full crash detail and a working theory connecting it to the stale window-designer error icons in #142.

Shipping this fix anyway because it's a real, well-understood bug on its own terms (a documented WinForms/COM gotcha, verifiable independent of the native crash), and closing the race is worth doing regardless of whether #179 turns out to be the dominant cause of what users are seeing.

## Test plan

- [x] Reproduced pre-fix: closing an embed tab immediately after typing (diagnostics request in flight) crashed reliably.
- [x] Post-fix: repeated close cycles in the same pattern, no crash from this specific path.
- [ ] Not fully conclusive on its own — see #179, which reproduced independently after this fix was live.

### DIFF
--- a/ClarionAssistant/Terminal/MonacoEditorControl.cs
+++ b/ClarionAssistant/Terminal/MonacoEditorControl.cs
@@ -355,24 +355,40 @@ namespace ClarionAssistant.Terminal
 
         /// <summary>Send a JSON message host-&gt;page (PostWebMessageAsJson), marshalled to the UI
         /// thread. Note: WebView2 delivers it to JS as a parsed OBJECT, not a string — page handlers
-        /// must not JSON.parse it. Used by the host for the state-assembled messages (SendSource etc.).</summary>
+        /// must not JSON.parse it. Used by the host for the state-assembled messages (SendSource etc.).
+        ///
+        /// GPF-on-close guard: a background continuation (e.g. the diagnostics settle-loop in
+        /// ModernEmbeditorDiagnostics.ComputeAsync, which can now run for several seconds) can still be
+        /// in flight when the embed is closed and DetachOverlay()/Dispose() tears this control down.
+        /// `_webView != null` alone does NOT detect that — Dispose() never nulls the field, only IsDisposed
+        /// flips. Worse, `InvokeRequired` on a control whose handle was just destroyed (mid-teardown, or
+        /// its parent chain was just severed by Controls.Remove) returns FALSE — not because we're on the
+        /// UI thread, but because it has no handle left to compare against — which used to fall through to
+        /// calling the WebView2/CoreWebView2 COM object (STA) directly from a thread-pool thread instead of
+        /// marshalling. That's undefined behaviour for an STA COM object mid-teardown and can produce a
+        /// native access violation no C# try/catch can intercept. IsDisposed is checked before trusting
+        /// InvokeRequired at all, and again inside the marshalled action (the two checks can't be merged
+        /// into one atomic test, but this closes the window down to the marshalling call itself).</summary>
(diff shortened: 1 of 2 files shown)


ClarionLive:
Merged as f4269a0 — thanks, this is a good catch.

The `InvokeRequired`-returns-false-on-a-destroyed-handle detail is the part that makes this worth fixing at the chokepoint rather than at each call site, and `PostJson`/`PostString` were indeed still unguarded on `master`.

Resolution note: the README needed a manual merge. `master` had grown its own `## What's New (Unreleased)` section in the meantime (VS Code settings import, build tools, templates, tool-list audit), so I kept those entries and appended yours below them under a `<!-- release-docs: covered=embeditor-post-dispose -->` marker to match the convention the other entries use. Your `---` separator before `## What's New in v5.6` was kept too — every other version section already has one, so that was a real fix.

Verified with a clean Debug build (0 errors) before merging.

Agreed that this doesn't close #179 — that stays open as the native Generator AV, along with #142.

ClarionLive:
Shipped in **[5.7](https://github.com/ClarionLive/ClarionAssistant/releases/tag/v5.7.0)** — thanks Adrián.

The `InvokeRequired` half of this is the part worth other people reading: on a control whose window handle has just been destroyed it returns `false`, not because the caller is already on the UI thread but because there's no handle left to compare against — so the post falls straight through to calling an STA COM object from a thread-pool thread mid-teardown. A null check on the field doesn't catch it either, since `Dispose()` never nulls it. Putting the guard at the single chokepoint every host→page message funnels through was the right place for it.

As the release notes say, this closes a real and distinct race but isn't the whole embed-close story: the native Access Violation in Clarion's Generator you reported as #179 reproduced after this was deployed and is still open, listed under Known issues in 5.7. It's likely the same underlying step behind the stale window-designer error icons in #142.

Also worth saying — the Ctrl+J snippet picker you originally asked for in #49 and #154 has moved to **Ctrl+Shift+J** in this release, because the Clarion embeditor uses Ctrl+J for next-filled-embed (#185). It's rebindable now, so it can go straight back to Ctrl+J if you prefer it there.

PR #182, opened 2026-08-07
Fix/codegraph definition local var guard

geircodes:
## Summary

F12/Ctrl+Click's CodeGraph fallback (`CgDefinitionFromDb` in `SharedLspBridge.cs`) could jump to
an unrelated procedure's local variable or parameter instead of correctly reporting "no
definition" — arbitrary and index-order-dependent, so the target could change across re-indexes
even though the underlying code hadn't changed.

## Root cause

`GetDefinition` asks the upstream Clarion LSP first; when it returns nothing — which mostly
happens when a word is genuinely undeclared, exactly the case where landing anywhere is wrong —
it falls back to `CgDefinitionFromDb`, a flat, unordered exact-name lookup against the indexed
CodeGraph database (`CodeGraphProvider.FindSymbolByName`, `WHERE LOWER(name)=LOWER(@name) LIMIT
1`). No `ORDER BY`, no scope filter. If the only matches anywhere in the solution are
procedure-local variables or parameters belonging to other, unrelated procedures, F12 still jumps
to one of them.

The correct guard already existed and was already used by the sibling hover path
(`CgHoverFromDb`, same file):

```csharp
private static bool IsUnreachableLocalVariable(CodeGraphProvider p, CodeGraphSymbol sym)
{
    if (!string.Equals(sym.Type, "variable", StringComparison.OrdinalIgnoreCase)) return false;
    if (string.IsNullOrEmpty(sym.ParentName)) return false;
    var parent = p.FindSymbolByName(sym.ParentName);
    return parent != null &&
        (string.Equals(parent.Type, "procedure", StringComparison.OrdinalIgnoreCase) ||
         string.Equals(parent.Type, "routine", StringComparison.OrdinalIgnoreCase));
}
```

`CgDefinitionFromDb` never called it — a missing sibling application, not a design choice. A
genuine module/global-scope variable (empty `ParentName`) is unaffected either way, since the
guard short-circuits to `false` before it would ever be filtered.

## Fix

`CgDefinitionFromDb` now mirrors `CgHoverFromDb`: fetches the full `CodeGraphSymbol` via
`FindSymbolByName` (instead of the thinner `GetDefinition` helper it used before) and rejects the
match when `IsUnreachableLocalVariable` says so.

## Verification

- Clean build, no new warnings (`ClarionAssistant.csproj`, VS2022 MSBuild toolset).
- Live-tested against the real trigger that motivated this fix: a call argument left undeclared
  by mistake, whose name happened to already be a local variable declared in a different method
  of the same class. Before the fix, F12 on the undeclared argument jumped to that unrelated
  method's own local declaration. After the fix, it correctly finds nothing. A second undeclared
  argument of a `QUEUE` type, in the same call, also correctly finds nothing after the fix (F12
  and Ctrl+Click both) — not independently tested before the fix, but it has the identical
  single-candidate DB shape as the first case, which was.
- Added `indexer/test-fixtures/codegraph-repro`'s "Bug Q" section: the real repro (a `QUEUE,TYPE`
  used as an optional method parameter, with the undeclared arguments' names shared with locals
  declared elsewhere in the class) as permanent regression coverage, plus two SQL queries that
  simulate `FindSymbolByName`'s exact lookup shape against a freshly-indexed DB and confirm the
  picked row's parent resolves to `type='procedure'` — i.e. exactly what `IsUnreachableLocalVariable`
  needs to see to reject it.
- No live-reproducible over-rejection case exists to test against: the guard only ever rejects a
  symbol that is `type='variable'` with a parent resolving to `procedure`/`routine`, and by
  Clarion's own scoping rules that exact shape is never legitimately reachable from outside its
  own owning procedure. Two independent attempts to find a live counter-example (a bare ABC class
  name, a bare structural keyword) both turned out to resolve through the upstream Clarion LSP's
  own provider rather than this fallback at all, which is consistent with the fallback's fix
  correctly having no legitimate case in its rejection condition.

## Related, not fixed here

While testing this fix's real-world trigger, a second, independent bug was found in
Clarion-Extension's own hover/definition resolver (`MemberLocatorService.isVariableLookupCandidate`):
it guards against resolving a bare word into a CLASS/INTERFACE member
(`context.inClass || context.inInterface`) but has no equivalent check for
`context.inQueueOrGroupOrRecord`, even though `StructureContext` already exposes that flag. A
QUEUE/GROUP/RECORD field has the same property a CLASS member does — only reachable via qualified
access, never as a bare word. Filed upstream separately; unrelated codebase, out of scope here.

### DIFF
--- a/ClarionAssistant/Services/SharedLspBridge.cs
+++ b/ClarionAssistant/Services/SharedLspBridge.cs
@@ -1409,8 +1409,15 @@ namespace ClarionAssistant.Services
             return null;
         }
 
-        /// <summary>Resolve an exact-name definition from one CodeGraph-schema DB → an LSP location list,
-        /// or null when the DB is missing/unopenable or the symbol isn't found. Never throws.</summary>
+        /// <summary>Resolve an exact-name definition from one CodeGraph-schema DB → an LSP location list, or
+        /// null when the DB is missing/unopenable, the symbol isn't found, or the only match is a "variable"
+        /// scoped to a PROCEDURE/ROUTINE elsewhere (see IsUnreachableLocalVariable — never a legitimate hit,
+        /// since the caller only reaches here after the upstream LSP already searched this file's own
+        /// local/routine/module scope and found nothing). Mirrors CgHoverFromDb's guard — that sibling
+        /// already filters this exact shape for hover; this path never got the same check, so a bare word
+        /// genuinely undeclared in scope (LSP correctly returns empty) could still resolve F12 to an
+        /// unrelated procedure's local via CodeGraphProvider.FindSymbolByName's unordered `LIMIT 1`. Never
+        /// throws.</summary>
         private static Dictionary<string, object> CgDefinitionFromDb(string word, string db)
         {
             try
@@ -1419,9 +1426,10 @@ namespace ClarionAssistant.Services
                 using (var p = new CodeGraphProvider())
                 {
                     if (!p.Open(db)) return null;
-                    var def = p.GetDefinition(word);
-                    if (def == null || string.IsNullOrEmpty(def.FilePath)) return null;
-                    return WrapResult(new System.Collections.ArrayList { CgLocation(def.FilePath, def.LineNumber) });
+                    var sym = p.FindSymbolByName(word);
+                    if (sym == null || string.IsNullOrEmpty(sym.FilePath)) return null;
+                    if (IsUnreachableLocalVariable(p, sym)) return null;
+                    return WrapResult(new System.Collections.ArrayList { CgLocation(sym.FilePath, sym.LineNumber) });
                 }
             }
             catch { return null; }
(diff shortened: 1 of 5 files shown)


ClarionLive:
Shipped in **[5.7](https://github.com/ClarionLive/ClarionAssistant/releases/tag/v5.7.0)** — thanks @geircodes.

Reviewed it against the source rather than the description, and the description held up: `CodeGraphProvider.GetDefinition(name)` is literally a wrapper around `FindSymbolByName(name)` projected to `{FilePath, LineNumber}`, so the swap loses nothing and only retains the full symbol for the guard. `IsUnreachableLocalVariable` is the existing shipping guard and `CgHoverFromDb` applies it with the identical three lines — a missing sibling application, exactly as you framed it.

One limitation recorded rather than left implicit: unlike hover, the definition path has no `BufferLocalHover`-style buffer-local pre-pass, so with the language server unavailable, F12 on a genuine in-file local now reports "no definition" where it previously jumped. That's the right trade — the old behaviour was an unordered `LIMIT 1`, a coin flip that could just as easily land on another procedure's local — but it's tracked so the pre-pass can be added properly.

The second bug you documented alongside this one has gone upstream as [Clarion-Extension issue 396](https://github.com/msarson/Clarion-Extension/issues/396)'s neighbour: `MemberLocatorService.isVariableLookupCandidate` guards `inClass || inInterface` but has no equivalent for `inQueueOrGroupOrRecord`, even though `StructureContext` already exposes the flag — so a QUEUE field resolves as a bare word. Filed on our side to send as a PR to Mark.

The fixture case was appreciated too. Deliberately non-compiling is the right call there: referencing an undeclared identifier *is* a hard Clarion compile error, which is the whole point, and the indexer parses text regardless.

Issue #126, opened 2026-07-18
Editor/formatter indentation should follow Clarion's Options (Indentation size, Tab size, Convert Tabs to Spaces) instead of assuming 4


ClarionLive:
## Context

Split from #70 (BoxSoft's tangent at the end): CA's editor surfaces currently assume 4-character indentation, but Clarion already knows the developer's preference — **Options → Text Editor → Behavior** has `Indentation size`, `Tab size`, and `Convert Tabs to Spaces`. We should read those instead of assuming.

## Scope

1. **Read the three settings from the IDE** via the fork's PropertyService (the same store the native editor reads; resolve the exact property keys by reflection on a live IDE — they live under the ICSharpCode text-editor property bundle).
2. **Pass them through on LSP formatting requests** as standard `FormattingOptions` (`tabSize`, `insertSpaces`). The bundled v1.0.0 server's `ClarionFormatter` already honors `tabSize` (constructor, ClarionFormatter.ts) — today it just never receives the real value.
3. **Align Monaco editor options** in the CA Editor / CA Embeditor (`tabSize`, `insertSpaces`, `detectIndentation: false`) with the same values, so what you type matches what formatting produces.
4. Re-read on solution open (settings are per-IDE, not per-solution — cheap to refresh at the same point the LSP restarts).

## Non-goals

Keyword casing — that's the main body of #70, home is the upstream formatter (msarson/Clarion-Extension#375), with the per-solution settings file tracked in #70 itself.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

peterparker57:
Implemented on master (58f7a39) and verified in-IDE. The CA editors and the Smart Formatter now follow **Options → Text Editor → Behavior** — Indentation size, Tab size, and Convert Tabs to Spaces — instead of assuming a fixed width. Details:

- A new **"Follow Clarion's editor options"** checkbox in the editor gear panel, ON by default. While on, the manual Tab size / Insert spaces settings are grayed but preserved, so unticking restores your own values exactly.
- Monaco has a single `tabSize`, so the mapping picks the value that governs what lands in the buffer: *Indentation size* when converting tabs to spaces, *Tab size* when inserting real tabs.
- Changes propagate **live**: OK the Options dialog and any open CA editor/embeditor tabs pick the new values up immediately.
- `detectIndentation` is pinned off — indentation is driven by configuration, never guessed from file content.

[...]

PR #183, opened 2026-08-07
fix(editor): follow Clarion's live font, not a dead property

geircodes:
## Summary

"Follow Clarion's editor options" in the Monaco gear panel (`Terminal/monaco-embeditor.html` + `Services/IdeEditorOptions.cs`) is supposed to mirror the IDE's own Options → Text Editor settings live. Font never actually worked: it read `TextEditorSettings.DefaultFont`, a property the current Options dialog doesn't write to — changing the font in Options never changed anything Monaco used, no matter how many times you toggled the checkbox.

## Root cause + fix

`IdeEditorOptions.ParseFont` parsed `TextEditorSettings.DefaultFont` (`[Font: Name=..., Size=..., Units=...]`). Verified directly against `ClarionProperties.xml`: changing the font size in Options → Text Editor and clicking OK rewrites the file, but `DefaultFont`'s `Size` never changes — it's dead. The value the dialog actually writes, and that the native IDE editor windows actually use, lives in a different bundle: `CoreProperties.ComponentsFont`'s `ListOfFonts` array, one pipe-delimited entry per IDE component (`"TextEditor|Text Editor, Output Window (Proportional Font)|Courier New,20"`).

Added `IdeEditorOptions.ParseComponentFont`, which finds the `TextEditor` entry and parses the trailing `Name,Size` pair (points, same pt→px conversion `DefaultFont` already used). Falls back to the old `DefaultFont` parse only if the new bundle is absent (older installs).

## Two more bugs in the same follow-mode gear panel, found while fixing the above

- **Font size/family never grayed out or showed the effective IDE value while following was on.** `syncFollowIdeUi()`/`populateSettingsPanel()` had a conditional disable + "show the effective state" branch for Cursor-behind-EOL, but nothing at all for font — it always showed/used the stored pref regardless of follow-mode. Added the missing branches, mirroring the existing Cursor-behind-EOL pattern.
- **Unchecking "follow" clobbered the stored font pref.** While following, the font fields display the *effective* (IDE) value but stay tied to the same stored-pref save path. `setFollowIdeIndent`'s `change` listener saved via `readSettingsPanel()`, which reads straight off the DOM — so unchecking, with the fields still showing the IDE's font, immediately persisted that IDE value as if it were the user's own preference, permanently overwriting whatever they'd actually set. Fixed by restoring the stored font/family/cursor-behind-EOL values into the fields *before* the save fires, only when transitioning to unchecked.

## Also fixed: live IDE font changes didn't propagate while following

`MonacoSettingsBroadcaster`'s `PropertyService.PropertyChanged` hook auto-re-broadcasts settings to every open Monaco surface when the IDE's Options dialog is OK'd — but only watched the `"TextEditorSettings"` key (GH #126's original scope, tab size/indentation). Font lives under `"CoreProperties.ComponentsFont"`, a different bundle the hook didn't know about, so an IDE font change while following required a manual uncheck/recheck to show up. Added `"CoreProperties.ComponentsFont"` to the watched keys.

## Verification

All four fixes tested live against a real Clarion 11 IDE session, iterating directly against `ClarionProperties.xml` and a temporary diagnostic build (raw `ComponentsFont` dump + on-page `_ideOpts`/disabled-state readout, both stripped before this commit) to confirm each piece independently before removing the scaffolding:
- Live IDE font size/family changes (`10pt`→`14pt`, confirmed via the raw property) correctly convert and apply while following.
- Font size/family fields correctly gray out and display the effective IDE value while checked.
- Unchecking correctly restores the stored pref (tested: stored "JetBrains Mono / 12" survives check→uncheck against a live "Courier New / 19" IDE font, in both directions, repeatedly).
- `settings.txt` persists correctly on every toggle — no clobbering in either direction.
- IDE font changes while following now propagate immediately, no manual toggle needed.

3 files changed, 104 insertions / 12 deletions on top of `upstream/master`.
check/recheck.

### DIFF
--- a/ClarionAssistant/Services/IdeEditorOptions.cs
+++ b/ClarionAssistant/Services/IdeEditorOptions.cs
@@ -132,9 +132,15 @@ namespace ClarionAssistant.Services
                 string.Equals((GetString(tes, "LineViewerStyle", "None") ?? "").Trim(), "None",
                               StringComparison.OrdinalIgnoreCase) ? "none" : "line";
 
-            // --- Font. Parse failures leave FontFamily=""/FontSize=0 so the caller falls back to the stored
-            // pref; we never substitute a font of our own choosing. ---
-            ParseFont(GetString(tes, "DefaultFont", null), v);
+            // --- Font. NOT read from the "TextEditorSettings" bundle above: "DefaultFont" there is a
+            // separate, dead property that the live Options dialog does not write to — verified directly
+            // against ClarionProperties.xml (changing font size in Options -> Text Editor updates
+            // ComponentsFont's "TextEditor" entry below, DefaultFont's Size stays wherever it started).
+            // Parse failures leave FontFamily=""/FontSize=0 so the caller falls back to the stored pref; we
+            // never substitute a font of our own choosing. ---
+            ParseComponentFont(v);
+            if (v.FontFamily.Length == 0 && v.FontSize == 0)
+                ParseFont(GetString(tes, "DefaultFont", null), v);   // older installs without the bundle below
 
             values = v;
             return true;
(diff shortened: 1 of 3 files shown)


ClarionLive:
Shipped in **[5.7](https://github.com/ClarionLive/ClarionAssistant/releases/tag/v5.7.0)** — thanks @geircodes.

The root cause here was a good catch, and not one that would have been found by reading the code: `TextEditorSettings.DefaultFont` *looks* like the right property, and the only way to know the Options dialog no longer writes to it is to change the font and watch `ClarionProperties.xml`. Three fixes in one PR — the live `ComponentsFont` bundle, the fields not graying out while following, and the broadcast hook not watching the font bundle so an IDE change needed a manual uncheck/recheck.

One follow-up went in before release, flagging it so it's on the record rather than a silent edit on top of your work. Making the fields display the IDE's *effective* value opened a path to losing the developer's own font: `readSettingsPanel()` reads the font straight off the DOM when saving, and a disabled input's `.value` still reads fine — so with following on, toggling any unrelated setting (minimap, word wrap, a formatter option) persisted the IDE's font as the stored preference, permanently. It also defeated the uncheck-restore in the PR, because `lastSettings` is reassigned from every locally-built payload and so tracked the panel rather than the preference.

Fixed with a `storedPref()` helper — the save-side inverse of `effIde()` — plus a stored-prefs snapshot captured only from host payloads. Cursor-behind-EOL had the same shape (since #126) and is fixed with it, and a VS Code import still counts as a deliberate new preference. Verified with a harness driving the real extracted functions: 8 assertions, which fail on the pre-fix source at exactly the three clobber cases and pass the four controls in both.

Issue #77, opened 2026-07-12
Clarion-Extension v1.0.0 released — bundled LSP pin (v0.9.8) is two releases behind; recommend re-pin


msarson:
Heads-up in the spirit of #40/#51: **msarson/Clarion-Extension hit 1.0.0 on 2026-07-11** (v0.9.9 shipped shortly before). The bundled pure LSP is currently pinned at **v0.9.8** (`lsp-snapshot.json`), so the bundle is two stable releases behind. Your own tooling makes the catch-up a one-liner:

```powershell
.\lsp-server-sync\Sync-LspServer.ps1 -Apply -Tag v1.0.0
```

## The headline: solution loading — speed and correctness

The bulk of the 1.0.0 release was a startup/solution-loading campaign, and it matters to this addin's server just as much as to VS Code:

- **Redirection sections for the active configuration were silently dropped** (Clarion-Extension #293): the build configuration arrives as `"Config|Platform"` but redirection sections are named `[Debug]`/`[Release]`/custom, and the strict comparison matched neither — only `[Common]` entries survived, so per-configuration generated-source directories (`genfiles\src` etc.) vanished from the search paths and **~99% of a real solution's source files failed to resolve at load** (41 of 3,016 measured). Note the CA addin sends `configuration=Release` in `clarion/updatePaths` — a v0.9.8 bundle is exposed to exactly this on redirection-heavy solutions. Downstream, cross-MEMBER navigation and the file-relationship graph were silently operating on ~1% of the solution.
- **Startup responsiveness overhaul** (#297): one sequenced, time-sliced startup pipeline; background indexing can no longer starve interactive requests. Measured on a 40-project / 3,000-file solution: solution ready in ~2s, interactive in ~5s (previously minutes).
- **No more multi-second event-loop freezes starving hover/completion** — the embeditor's exact request pattern: first cross-file validation froze ~8s (#319, now 1.8s cold with 42 instead of 3,055 file tokenizations), CodeLens resolves could block 106s+ (#303/#309), SELF/PARENT hover walked include chains cold for 2–12s (#314), cursor-rest on an INCLUDE line cost up to 5.7s per move (#312).
- **Indexes persist across sessions** (#290/#307/#311): structure index and MODULE/INCLUDE/MEMBER file graph warm-start from mtime-keyed caches instead of rebuilding (~27s → roughly the changed files only).
[...]

Issue #42, opened 2026-06-08
better-sqlite3 missing from bundled lsp-server/node_modules — CodeGraphBridge self-disables on end-user installs


peterparker57:
@
Splitting this out from #30 as promised, so the packaging fix doesnt get lost in the broader contributor-onboarding thread.

## Problem

The `lsp-server/node_modules/` shipped by the installer contains only the LSP-protocol packages in `deploy.ps1`s `$LspNodeModules` allowlist — **`better-sqlite3` is not included**. As a result, the LSP servers CodeGraphBridge self-disables on end-user installs:

```
[CodeGraphBridge] ❌ ERROR: better-sqlite3 not available — CodeGraph features disabled
```

Core LSP features (`definition` / `references` / `hover` / `diagnostics` / `rename`) still work; only the CodeGraph bridge *inside the LSP server* is affected.

Confirmed against the installed addin by @dinkobukic (see #30).

## Fix options

1. Add `better-sqlite3` (and its transitive native deps) to `deploy.ps1`s `$LspNodeModules` allowlist, **and** ensure the bundled prebuilt native module matches the bundled `node.exe` ABI/version.
2. Or bundle a prebuilt `better-sqlite3` binary matching the shipped Node runtime.

The ABI match is the subtle part — a `better-sqlite3` built for a different Node version will fail to load even if copied.

## Acceptance

- Fresh install (no dev environment) boots the LSP with CodeGraphBridge **enabled** (no "better-sqlite3 not available" error).
- `query_codegraph` works against the installer-bundled LSP, not just a from-source deploy.

Related: #30 (contributor onboarding / LSP source), #40 (LSP snapshot drift).
@

peterparker57:
**Verified resolved — closing.**

The packaging fix has landed in both build paths:
- `ClarionAssistant/deploy.ps1` — `$LspNodeModules` now includes `better-sqlite3`, `bindings`, `file-uri-to-path`
- `installer/ClarionAssistant.iss` — those three modules are bundled for **C10/C11/C12** (lines 144–146 / 190–192 / 236–238)

**ABI match confirmed against the actual shipped artifacts** (not just "it's in the list"). Bundled `node.exe` is v24.13.0, better-sqlite3 is 12.8.0, and loading the bundled native module under the bundled node succeeds:

```
node v24.13.0 ABI 137
LOAD OK — CodeGraphBridge dependency works
```

[...]

PR #186, opened 2026-08-08
fix(lsp): make -Pure self-contained and verify the shipped LSP matches the pin

bdinko:
# fix(lsp): make `-Pure` self-contained and verify the shipped LSP matches the pin

## Why

The bundled-LSP pin is the only thing a consumer can inspect to know which language server they
received. Three defects let "what shipped" drift from "what `lsp-snapshot.json` says shipped" —
and one of them makes a pin bump abort the whole deploy on any machine that is not the
maintainer's.

I hit the consumer-facing end of this in early July: I synced ClarionAssistant, deployed, and
LSP-backed features did not behave as demonstrated. At that point the pin was `v0.9.8`. The
subsequent re-pin commit `ed8bde5` (`fix(lsp): re-pin bundled LSP to Clarion-Extension v1.0.0
(#77)`) describes exactly what I was seeing:

> pulls in upstream's redirection-section fix (per-config sections were silently dropped
> — CA sends configuration=Release, so redirection-heavy solutions resolved ~1% of source
> files), the startup/freeze overhaul, persisted indexes, and regeneration-burst coalescing.

On a redirection-heavy solution the server resolved about **1% of source files** while appearing
installed and healthy. Nothing available to me distinguished "I am using it wrong" from "the
bundled LSP predates the fix". The same commit notes the mirror-image problem on the producing
side — *"dev-machine tests were masked by a stray C:\Users\<user>\node_modules"*. Both directions
have one root cause: the LSP that ships is never verified against the LSP that is pinned.

### 1. `-Pure` no longer needs the maintainer's clone

`$LspRoot` defaulted to `H:\DevLaptop\ClarionLSP` and was validated at the top of the script,
*before* the `-Pure` block could run. The pure tree was then created with
`git worktree add --force $PureRoot $Tag` — a worktree **of that clone**. So `-Pure` was
unreachable on any other machine, despite its own comment describing it as "Self-contained".

The knock-on effect is the serious one. `Resolve-LspBuild` calls `-Pure` whenever
`.lsp-build\<tag>\out\server\src\server.js` is missing, which is exactly what a pin bump causes.
On a contributor machine that call dies, and because `deploy.ps1` sets
`$ErrorActionPreference = "Stop"`, the whole deploy dies with it — before a single Clarion version
builds. That defeats the soft-skip the caller clearly intends:

```powershell
deploy.ps1:205  if ($LASTEXITCODE -ne 0) { "WARN pure LSP build failed ... LSP copy will be skipped." }
```

The outcome was not "deploy without an LSP". It was no deploy at all.

`-Pure` now runs **before** any `$LspRoot` resolution and clones `$manifest.source.repo` at the
tag itself — the URL was already in the manifest. It needs only git, npm and network. `$LspRoot`
is now used solely by the legacy overlay path (`-Apply`), and its doc comment says so.

### 2. A missing drive now fails soft instead of throwing

`Join-Path` validates the drive qualifier and *throws* rather than returning false:

```
PS> $ErrorActionPreference='Stop'; Join-Path 'H:\DevLaptop\ClarionLSP' '.git'
Cannot find drive. A drive with the name 'H' does not exist.
```

`Test-Path` alone returns `$false`, so the check is now
`Test-Path -LiteralPath "$LspRoot\.git"`.

### 3. `deploy.ps1` verifies the LSP against the pin before copying

New `Assert-LspPin` compares the source tree's `HEAD` to `resolvedCommit` and exits 7 on mismatch,
so a tree reached via `$env:CLARIONLSP_ROOT` — or any `out/` built from a different tag — cannot
silently ship under a pin that describes something else. Prefix comparison, since
`git rev-parse --short` auto-scales its length.

`CLARIONLSP_ROOT` remains available as the dev escape hatch, but now announces itself:

```
WARN  CLARIONLSP_ROOT is set — shipping an UNPINNED LSP from <path>
WARN  lsp-snapshot.json will NOT describe what ships. Do not use this for a release build.
```

and downgrades the mismatch to a warning when it is in effect, on the basis that it is a
deliberate override.

### 4. (Fallout) The manifest round-trip corrupted the file on PowerShell 5.1

Not a pre-existing symptom for anyone, because nobody outside the maintainer's machine could run
`-Pure` — fixing (1) makes it reachable for every 5.1 contributor, so it is fixed here rather
than left as a new regression path.

`Get-Content -Raw` defaults to the system ANSI codepage in Windows PowerShell 5.1 and
`Set-Content -Encoding UTF8` writes UTF-8 **with** BOM. A read/write round-trip therefore turned
the em dash in `$comment` into `â€”` and prepended a BOM; `ConvertTo-Json` additionally escaped
`&`, `<`, `>`, `'` as `\uXXXX` (5.1 has no `-EscapeHandling`). PS 7 defaults to UTF-8, which is
why this went unnoticed.

`Read-Manifest` / `Save-Manifest` now round-trip losslessly. **Known limitation:** indentation
still reflows to 5.1's `ConvertTo-Json` alignment style, so a 5.1 contributor's sync produces a
whitespace-only reflow of the file. Content is unaffected. Happy to add a canonical formatter if
you want the churn gone.

## Testing

Run on a machine that has **never** had the `H:` clone — the configuration that surfaces all of
these:

| Test | Result |
|---|---|
| `-Pure` with no `H:` drive, fresh `PureRoot` | Clones `source.repo`, checks out `v1.0.0`, asserts no overlay, builds, resolves `571ffcee` — matches the pin |
| `-Pure` again on the same tree | Refreshes checkout, skips rebuild (idempotent), still `571ffcee` |
| Legacy path (no `-Pure`) with no `H:` drive | `FAIL Not a git clone: H:\DevLaptop\ClarionLSP`, exit 2 — no `Join-Path` throw |
| `deploy.ps1 -Version 12 -NoBuild` | `OK bundled LSP matches pin: 571ffcee (v1.0.0)`, deploy completes, exit 0 |
| Same, with `resolvedCommit` forced to a bogus value | `FAIL bundled LSP is 571ffcee but lsp-snapshot.json pins deadbeef`, exit 7, nothing deployed |
| Manifest round-trip | No BOM, em dash intact, `&&` and `'` literal, valid JSON, values unchanged |
| Both scripts | Parse clean (`[Parser]::ParseFile`) |

Full four-version deploy (C10 / C11.0.13505 / C11.1-13810 / C12) also verified working with these
changes on the current pin.

## Notes

- No behavioural change for anyone who already has the `H:` clone — the legacy `-Apply` path is
  untouched apart from where its `$LspRoot` validation lives.
- `lsp-snapshot.json` is deliberately **not** modified in this PR; only the two scripts change.
- Happy to adjust the exit code, or to make the pin mismatch a warning rather than a hard failure,
  if you would rather not have it block a deploy.

---

### DIFF
--- a/ClarionAssistant/deploy.ps1
+++ b/ClarionAssistant/deploy.ps1
@@ -277,32 +277,41 @@ function Invoke-GitQuiet([string[]]$GitArgs) {
     }
 }
 
-function Assert-LspPin($sourceDir) {
+# Returns $true when the LSP may ship, $false when it must not. It does NOT exit: an LSP-copy
+# problem is not a reason to abandon the Clarion builds. This used to `exit 7`, which aborted all
+# four Clarion targets AND the C# addin before a single one was built -- cutting straight across
+# the "a build failure for ONE Clarion release must not block shipping to the others" rule the
+# final summary below is built around. The caller skips just the LSP copy instead, so every version
+# still deploys and keeps whatever server it already had, and the run ends non-zero so nobody reads
+# it as clean.
+function Test-LspPin($sourceDir) {
     $manifest = Get-Content (Join-Path $ProjectDir "lsp-server-sync\lsp-snapshot.json") -Raw | ConvertFrom-Json
     $pinned   = $manifest.resolvedCommit
     if (-not $pinned) {
         Write-Host "  WARN  manifest has no resolvedCommit — cannot verify the bundled LSP." -ForegroundColor Yellow
-        return
+        return $true
     }
     $head = Invoke-GitQuiet @('-C', $sourceDir, 'rev-parse', '--short', 'HEAD')
     if (-not $head) {
         Write-Host "  WARN  $sourceDir is not a git tree — cannot verify it matches pin $pinned." -ForegroundColor Yellow
-        return
+        return $true
     }
     # Prefix compare: `git rev-parse --short` auto-scales its length, so 7- and 8-char forms of the
     # same commit must still count as a match.
     if (-not ($head.StartsWith($pinned) -or $pinned.StartsWith($head))) {
         if ($env:CLARIONLSP_ROOT) {
             Write-Host "  WARN  bundled LSP is $head but the pin says $pinned (CLARIONLSP_ROOT override in effect)." -ForegroundColor Yellow
-            return
+            return $true
         }
         Write-Host "  FAIL  bundled LSP is $head but lsp-snapshot.json pins $pinned." -ForegroundColor Red
         Write-Host "        Re-run lsp-server-sync\Sync-LspServer.ps1 -Pure, or bump the pin deliberately." -ForegroundColor Red
-        exit 7
+        Write-Host "        The LSP copy will be SKIPPED for every version; the rest of the deploy continues." -ForegroundColor Red
+        return $false
     }
     Write-Host "  OK    bundled LSP matches pin: $pinned ($($manifest.resolvedTag))" -ForegroundColor Green
+    return $true
 }
-Assert-LspPin $LspSourceDir
+$LspPinOK = Test-LspPin $LspSourceDir
 # Pure v1.0.0 runtime deps only — NO better-sqlite3/bindings/file-uri-to-path (those backed the retired
 # CodeGraph overlay). With better-sqlite3 absent the #42 ABI check below self-skips ("module not deployed").
 # iconv-lite (+ its dep safer-buffer) is NEW at v1.0.0: UnicodeDiagnostics requires it in the EAGER
@@ -523,7 +532,7 @@ foreach ($ver in $TargetVersions) {
         # --- Deploy LSP Server ---
         $LspDestDir = Join-Path $DeployDir "lsp-server"
 
-        if (Test-Path $LspSourceDir) {
+        if ($LspPinOK -and (Test-Path $LspSourceDir)) {
             # Copy compiled server JS + common shared code
             foreach ($outDir in @("out\server", "out\common")) {
                 $LspOutSrc = "$LspSourceDir\$outDir"
@@ -611,6 +620,11 @@ foreach ($ver in $TargetVersions) {
             } elseif (Test-Path $DeployedNode) {
                 Write-Host "  SKIP  better-sqlite3 ABI check (module not deployed)" -ForegroundColor DarkGray
             }
+        } elseif (-not $LspPinOK) {
+            # Deliberate: shipping a server that does not match the pin is worse than shipping none,
+            # because lsp-snapshot.json would go on asserting a commit the consumer did not receive.
+            # The version keeps whatever lsp-server it already had; the final summary exits non-zero.
+            Write-Host "  SKIP  lsp-server (does not match the pin — see FAIL above)" -ForegroundColor Yellow
         } else {
             Write-Host "  SKIP  lsp-server (ClarionLSP not found)" -ForegroundColor DarkGray
         }
@@ -626,13 +640,19 @@ foreach ($ver in $TargetVersions) {
 
 # --- Final summary ---
 Write-Host ""
-if ($FailedBuilds.Count -gt 0) {
+if ($FailedBuilds.Count -gt 0 -or -not $LspPinOK) {
     # Never let a partial run exit 0 with a bare "All done." — that is what made a stale deployed DLL
-    # look like a successful deploy. Name the versions that did NOT ship, and fail the exit code so a
-    # caller (CI, the installer, another script) can't read this run as clean.
+    # look like a successful deploy. Name what did NOT ship, and fail the exit code so a caller (CI,
+    # the installer, another script) can't read this run as clean.
     Write-Host "Done, WITH FAILURES." -ForegroundColor Yellow
-    Write-Host "  NOT deployed (build failed): $($FailedBuilds -join ', ')" -ForegroundColor Red
-    Write-Host "  Every other requested version deployed normally." -ForegroundColor Yellow
+    if ($FailedBuilds.Count -gt 0) {
+        Write-Host "  NOT deployed (build failed): $($FailedBuilds -join ', ')" -ForegroundColor Red
+    }
+    if (-not $LspPinOK) {
+        Write-Host "  NOT deployed (pin mismatch): lsp-server — every version kept the LSP it already had." -ForegroundColor Red
+        Write-Host "  Re-run lsp-server-sync\Sync-LspServer.ps1 -Pure, or bump the pin, then re-run this script." -ForegroundColor Red
+    }
+    Write-Host "  Everything else deployed normally." -ForegroundColor Yellow
     exit 1
 }
 Write-Host "All done." -ForegroundColor Green
--- a/ClarionAssistant/deploy.ps1
+++ b/ClarionAssistant/deploy.ps1
@@ -242,6 +242,41 @@ $LspSourceDir = Resolve-LspBuild
 # server.js was copied (CLARIONLSP_ROOT override, hand-edited tree, an out/ built from another tag),
 # and the manifest is the ONLY thing a consumer can inspect to know what they got. Assert before any
 # copy so "what shipped" and "what is pinned" cannot silently diverge.
+
+# Run git and return its stdout, or $null if it failed for ANY reason. Do NOT replace this with a
+# bare `git ... 2>$null`.
+#
+# Under Windows PowerShell 5.1, redirecting a NATIVE command's stderr wraps the output in a
+# NativeCommandError record, and $ErrorActionPreference='Stop' (set at the top of this file) makes
+# that record TERMINATING. So `$x = (git ... 2>$null); if (-not $x) { warn; return }` never reaches
+# the soft-fail branch -- the script dies on the git line. Reproduced on 5.1.19041.6456 and on
+# 5.1.26100.8972; the identical code under pwsh 7 returns empty and continues, which is why the bug
+# looks like it isn't there when tested from a pwsh prompt.
+#
+# That is the exact failure this file's LSP handling is supposed to avoid: a contributor whose pure
+# build fails, or whose CLARIONLSP_ROOT is a plain unpacked server tree with no .git, would see the
+# WARN and then have the whole deploy abort anyway -- before a single Clarion version builds.
+#
+# Relaxing $ErrorActionPreference locally makes the NativeCommandError non-terminating; 2>&1 keeps
+# git's stderr off the console; the ErrorRecord filter keeps it out of the return value; the
+# try/catch covers git being absent from PATH entirely (CommandNotFoundException terminates
+# regardless of the preference).
+function Invoke-GitQuiet([string[]]$GitArgs) {
+    $prevEap = $ErrorActionPreference
+    $ErrorActionPreference = 'Continue'
+    try {
+        $out = & git @GitArgs 2>&1
+        if ($LASTEXITCODE -ne 0) { return $null }
+        $clean = $out | Where-Object { $_ -isnot [System.Management.Automation.ErrorRecord] }
+        if (-not $clean) { return $null }
+        return (($clean -join "`n").Trim())
+    } catch {
+        return $null
+    } finally {
+        $ErrorActionPreference = $prevEap
+    }
+}
+
 function Assert-LspPin($sourceDir) {
     $manifest = Get-Content (Join-Path $ProjectDir "lsp-server-sync\lsp-snapshot.json") -Raw | ConvertFrom-Json
     $pinned   = $manifest.resolvedCommit
@@ -249,7 +284,7 @@ function Assert-LspPin($sourceDir) {
         Write-Host "  WARN  manifest has no resolvedCommit — cannot verify the bundled LSP." -ForegroundColor Yellow
         return
     }
-    $head = (git -C $sourceDir rev-parse --short HEAD 2>$null)
+    $head = Invoke-GitQuiet @('-C', $sourceDir, 'rev-parse', '--short', 'HEAD')
     if (-not $head) {
         Write-Host "  WARN  $sourceDir is not a git tree — cannot verify it matches pin $pinned." -ForegroundColor Yellow
         return
--- a/ClarionAssistant/deploy.ps1
+++ b/ClarionAssistant/deploy.ps1
@@ -194,7 +194,14 @@ $Items = @(
 # tag checkout produced by lsp-server-sync\Sync-LspServer.ps1 -Pure, cached under .lsp-build\<tag>.
 # $env:CLARIONLSP_ROOT still overrides (dev escape hatch) if you deliberately want a custom server tree.
 function Resolve-LspBuild {
-    if ($env:CLARIONLSP_ROOT) { return $env:CLARIONLSP_ROOT }   # explicit override wins
+    if ($env:CLARIONLSP_ROOT) {
+        # Loud, because this is the one path that bypasses the tag-in-path pin check below: whatever
+        # is in that tree ships, while lsp-snapshot.json goes on asserting resolvedCommit. A consumer
+        # then has no way to tell which server they actually received.
+        Write-Host "  WARN  CLARIONLSP_ROOT is set — shipping an UNPINNED LSP from $env:CLARIONLSP_ROOT" -ForegroundColor Yellow
+        Write-Host "  WARN  lsp-snapshot.json will NOT describe what ships. Do not use this for a release build." -ForegroundColor Yellow
+        return $env:CLARIONLSP_ROOT
+    }
     $syncScript = Join-Path $ProjectDir "lsp-server-sync\Sync-LspServer.ps1"
     $manifest   = Get-Content (Join-Path $ProjectDir "lsp-server-sync\lsp-snapshot.json") -Raw | ConvertFrom-Json
     $tag        = if ($manifest.resolvedTag) { $manifest.resolvedTag } else { $manifest.targetPin.tag }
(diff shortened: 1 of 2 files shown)


ClarionLive (changes requested):
Thanks for this — the diagnosis is right and the writeup made it easy to verify. `-Pure` genuinely was unreachable off my machine, and you've correctly identified the serious knock-on: because `Resolve-LspBuild` calls it on any pin bump, and `deploy.ps1` runs under `$ErrorActionPreference = "Stop"`, a contributor's pin bump killed the entire deploy rather than taking the soft-skip the caller clearly intends. That's a real defect and I want the fix.

The `Join-Path` → `Test-Path -LiteralPath` change is correct, and I confirmed the manifest round-trip is genuinely lossless on 5.1 — no BOM, em dash intact, no residual `\uXXXX`, values compare `-ceq` to the originals. Don't worry about the indentation reflow; that's not worth chasing.

A few things to fix before this goes in.

## 1. Blocker — the new code reintroduces the same bug class it removes

`Assert-LspPin` does:

```powershell
$head = (git -C $sourceDir rev-parse --short HEAD 2>$null)
if (-not $head) { Write-Host "  WARN  ... cannot verify ..."; return }
```

Under Windows PowerShell 5.1, redirecting a **native** command's stderr turns the output into a `NativeCommandError` record, and `$ErrorActionPreference = "Stop"` makes that terminating. So that soft-fail `return` is unreachable — the script dies on the `git` line instead. Verified on 5.1.26100.8972: with `$sourceDir` set to a non-existent path the script terminated there and never reached the next statement, while the identical code under pwsh 7 returned empty and continued.

The effect is the failure mode this PR set out to remove. A contributor on `powershell.exe` whose pure build fails (offline, npm missing) sees `WARN pure LSP build failed — LSP copy will be skipped`, and then the deploy aborts anyway before a single Clarion version builds. It also fires when `CLARIONLSP_ROOT` points at a plain unpacked server tree with no `.git` — which is precisely the case the "not a git tree" warning was written for.

Two more instances of the same construct in this diff:
- `Sync-LspServer.ps1`, `$pureOrigin = (git -C $PureRoot remote get-url origin 2>$null)` — under that script's own `Stop`. Concrete trigger: a `.lsp-build/<tag>` left over from the old `git worktree add` implementation, whose `.git` file points into a parent clone that has since moved. The intended `Fail … exit 2` never runs; it throws, and since `deploy.ps1` invokes the script with `&` under its own `Stop`, that propagates and kills the deploy.
- the `rev-parse --short HEAD 2>$null` that resolves the commit for the manifest.

`Test-Path` guard, `try/catch`, or capturing with `2>&1` after locally relaxing `$ErrorActionPreference` all work.

## 2. Please rebase — master now has a fix you'll want

I've pushed 1ceb6f7 to master. `Resolve-LspBuild` was calling the sync script uncaptured inside a function that ends `return $pureDir` — so on the first-run build path, every line git/npm/tsc emitted became part of the return value and `$LspSourceDir` came back as an **array** with the real path merely last. It's now `& $syncScript -Pure -Tag $tag | Out-Host`.

This matters directly for your PR: `Assert-LspPin $LspSourceDir` is a new consumer of that value, and PowerShell would splat those output lines as separate arguments to `git -C`. Before your change the pollution was survivable — a silent `SKIP lsp-server` — but the new assert turns it fatal, and your `-Pure` restructure emits strictly more stdout down that path since it now clones and runs npm where it previously reused a local worktree. Rebasing onto master gets you a clean path for free; you'll take a small conflict in `Resolve-LspBuild` where the two changes touch adjacent lines.

## 3. The assertion doesn't check the artifact that ships

It compares the *source checkout's* `HEAD`, not the built `out/server/src/server.js` that actually gets copied — so it doesn't quite deliver what the comment above it claims about "an `out/` built from another tag". `out/` is gitignored and survives `git checkout --force`, and `-Pure` skips the rebuild when a `server.js` already exists. So `CLARIONLSP_ROOT` pointing at an overlay clone checked out at the pinned tag prints the green "matches pin" line while an overlay build ships. Same if an upstream tag is ever re-pointed: refresh moves HEAD, the rebuild is skipped, `resolvedCommit` is rewritten, and the assert agrees with itself.

Not a blocker — the check is still a strict improvement on nothing. But hashing the built `server.js` into the manifest is what would actually close the gap, and it'd be worth a follow-up.

## 4. Smaller things

- **Take your own offer on `exit 7`.** You asked whether it should block a deploy — it shouldn't. It aborts all four Clarion targets and the C# addin for what is an LSP-copy problem, and it fires before any build, which cuts against the "a build failure for ONE Clarion release must not block shipping to the others" design a bit further down the same file. A loud warning plus skipping the LSP copy, or a `-SkipLspPin` switch, would fit better.
- **`git fetch --tags --prune --quiet origin` has its exit status unchecked**, unlike the other git calls in that block. Offline on the refresh path, the fetch fails, `git tag --list` still finds the stale cached tag, the checkout "succeeds", the rebuild may be skipped, and the stale commit gets written into the manifest as authoritative — while the run prints `OK PURE sync complete` and your new pin check agrees.
- **`Save-Manifest`'s unescape runs over the serialized JSON text rather than over values.** If a note ever contains a literal backslash followed by `u0026` as ordinary text, `ConvertTo-Json` escapes the backslash, and the blanket unescape then collapses the result into a backslash-ampersand pair — an invalid JSON escape, so the file fails to re-parse. Nothing currently trips it, but `targetPin.note` and `$comment` are free-form prose maintained by hand. Validating with `ConvertFrom-Json` before writing would make it safe.

Three of these are the same root cause — native command exit/error handling under `Stop` — so it should be a contained round rather than a rewrite. Happy to look again as soon as you've pushed.

bdinko:
Thanks — that review was a pleasure to work from. The 5.1 repro in point 1 was the part I had no answer to, and you were right: I had tested that path from a pwsh prompt, where it looks correct.

Pushed three commits. The branch is rebased onto master first, so the diff you see is against `f3587fe`.

- `55d8b9a` — point 1, the blocker
- `147a030` — point 4, `exit 7` and the unchecked fetch
- `6ad2c4f` — point 4, `Save-Manifest`

---

## 1. `git ... 2>$null` under `Stop` — fixed, all three sites

Confirmed your diagnosis on 5.1.19041.6456 before changing anything. Pre-fix `Sync-LspServer.ps1 -Pure` against a `.lsp-build/<tag>` whose `.git` file points at a moved parent clone:

```
  INFO  Refreshing existing checkout -> v1.0.0
  OK    origin: https://127.0.0.1:1/msarson/Clarion-Extension.git
git : fatal: not a git repository: Z:\moved\away\clone\.git\worktrees\v1.0.0
    + CategoryInfo          : NotSpecified: (...) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
--- exit code: 1 ---
```

Exactly as you described — the `Fail … exit 2` never runs. Post-fix, same input, same shell: the intended `FAIL origin of … Aborting.` and exit 2.

All three sites now go through one helper, duplicated in both scripts since `deploy.ps1` doesn't dot-source the sync:

```powershell
function Invoke-GitQuiet([string[]]$GitArgs) {
    $prevEap = $ErrorActionPreference
    $ErrorActionPreference = 'Continue'
    try {
        $out = & git @GitArgs 2>&1
        if ($LASTEXITCODE -ne 0) { return $null }
        $clean = $out | Where-Object { $_ -isnot [System.Management.Automation.ErrorRecord] }
        if (-not $clean) { return $null }
        return (($clean -join "`n").Trim())
    } catch {
        return $null
    } finally {
        $ErrorActionPreference = $prevEap
    }
}
```

Relaxing the preference locally makes the `NativeCommandError` non-terminating, `2>&1` keeps git's stderr off the console, the `ErrorRecord` filter keeps it out of the return value, and the `try/catch` covers git missing from PATH entirely — `CommandNotFoundException` terminates regardless of the preference. It takes its args as a single array rather than `ValueFromRemainingArguments`, so a git flag can never be parsed as a parameter name.

One addition beyond the three sites: `$resolved` now warns when it comes back empty. That's a supported case — the "Using existing (non-git) pure tree as-is" branch — but silently writing `resolvedCommit: null` makes the deploy report "manifest has no resolvedCommit" with no hint why.

I verified the `deploy.ps1` end by extracting the real `Invoke-GitQuiet` and `Test-LspPin` out of the file via the AST and running them under the file's own `$ErrorActionPreference = "Stop"` — both the missing-path and the no-`.git` cases now warn and return instead of terminating.

## 2. Rebase — done, and `1ceb6f7` merged cleanly

Rebased onto `f3587fe`. The conflict you predicted in `Resolve-LspBuild` didn't materialise; git took both changes and I confirmed `& $syncScript -Pure -Tag $tag | Out-Host` survived intact. Thanks for the heads-up on that one — the splatting failure you describe would have been genuinely hard to find from the symptom.

## 3. The assertion checks the source, not the artifact — agreed, follow-up

You're right, and the comment above it does overclaim. `out/` being gitignored and surviving `git checkout --force`, plus `-Pure` skipping the rebuild when a `server.js` exists, means an overlay build can ship under a green "matches pin" line.

Opened as **#191**, draft and stacked on this branch since it modifies `Test-LspPin`. Two things worth flagging from writing it:

- **Hashing alone doesn't close your second case.** For a re-pointed tag, the sync would write the new commit *and* a hash of the old artifact — still self-consistent, still wrong. So #191 also stamps the commit that produced `out/` and rebuilds when HEAD has moved past it.
- **It fails open on an existing manifest**, so merging it verifies nothing by itself; the first `-Pure` run after the merge records the hash. There's a question in that PR for you about whether your `v1.0.0` build hashes the same as mine — if `npm ci` + `npm run compile` isn't byte-reproducible across machines, a shared hash field can't work and it needs rethinking before it merges.

## 4. Smaller things

**`exit 7` — taken, it no longer blocks.** `Assert-LspPin` is now `Test-LspPin` returning a bool (`Assert` implied a throw, and it no longer does). On a mismatch the LSP copy is skipped for every version, each keeps whatever `lsp-server` it already had, everything else deploys normally, and the final summary names it and exits 1.

I went with skip-the-copy rather than `-SkipLspPin` because shipping a server that doesn't match the pin seemed worse than shipping none — `lsp-snapshot.json` would go on asserting a commit the consumer didn't receive, which is the same drift the file is there to prevent. Happy to add the switch instead, or as well, if you'd rather have the escape hatch. The three benign paths are unchanged: no `resolvedCommit`, not a git tree, and the `CLARIONLSP_ROOT` override all still warn and ship. Nothing in the tree read exit 7.

The summary now handles both failure kinds independently, so a failed build and a pin mismatch each get named rather than one masking the other.

**`git fetch` exit status — checked.** Your offline chain is real, and the part I found most convincing is that it ends with the run printing `OK PURE sync complete` while writing a stale commit as authoritative. A tag that can't be confirmed against origin isn't a pin, so it now fails with exit 2 like the neighbouring clone/tag/checkout failures — which `deploy.ps1` already treats as the soft "LSP copy will be skipped" path rather than dying.

Verified by pointing a test clone's origin at a port nothing listens on:

```
  FAIL  fetch from origin failed (128) in … — cannot confirm 'v1.0.0' against upstream.
         A locally cached tag may be stale, so pinning the manifest to it would be a lie.
         Re-run with a working network connection.
--- exit code: 2 ---
```

Worth flagging one behaviour change: a hand-run `-Pure` while offline now fails where it previously appeared to succeed. I don't think that's a practical regression — `deploy.ps1` only invokes the sync when `out/server/src/server.js` is *missing*, so that path has to run `npm ci` anyway and would fail offline regardless — but it is a change, and if you'd rather have a loud WARN than a hard stop there, say so.

**`Save-Manifest` — parity-aware, and it validates before writing.** Confirmed your case is live on the old code:

```
  serialized: { "comment": "\&" }
  re-parse FAILED: Unrecognized escape sequence.
```

The unescape now decides by backslash parity: count the backslashes immediately before `uXXXX`; odd means the last one is a real escape introducer, so drop it and substitute; even means they all pair up into literal backslashes and `uXXXX` is ordinary prose, so leave the match alone. Then `ConvertFrom-Json` validates and it throws rather than writing — failing closed, so a bad serialization leaves the good manifest on disk and the error names the writer instead of surfacing as a `Read-Manifest` stack trace on someone else's next run.

11 round-trips through the real `Save-Manifest`/`Read-Manifest`, each comparing every leaf with `-ceq`: plain prose, all four escaped characters, an em dash, a Windows path, backslash + `u0026`, backslash + `u003c`/`u003e`/`u0027`, two and three backslashes before `u0026`, mixed literal and real ampersands, uppercase `U0026`, and tab + embedded quotes. All pass, none writes a BOM.

I also re-ran the property you checked, since I didn't want to trade one regression for another: the real `lsp-snapshot.json` still round-trips with zero value differences, no BOM, no residual `\uXXXX`, em dash intact.

---

## Two things I noticed but left alone

**`deploy.ps1:605`** — `$abiOut = & $DeployedNode -e $abiProbe 2>&1` is the same defect class as point 1, on a native command that isn't git. Pre-existing and outside this diff, but the consequence is that when the probe fails, node writes to stderr and the deploy terminates there — so `FAIL better-sqlite3 ABI mismatch (GitHub #42)` on the next line has, as far as I can tell, never actually printed. Say the word and I'll fix it here or separately.

**`deploy.ps1:207` vs `Sync-LspServer.ps1:183`** — the deploy resolves the tag as `resolvedTag ?? targetPin.tag`, while the sync defaults `-Tag` from `targetPin.tag`. A pin staged the way `targetPin.note` describes — hand-edit `targetPin.tag` to a newer release, commit, sync later — is therefore silently ignored by `deploy.ps1`, which keeps building the old `resolvedTag` and reports it as matching. Same family as point 3: the manifest can say one thing while the deploy uses another. Not touching it in this PR, but it seemed worth writing down.

`Sync-LspServer.ps1:321` and `:339` still use the old construct. Both are in the legacy `-Apply` path, which is hand-run and not in this diff, so I left them for the same "contained round" reason — happy to fold them in if you'd prefer the file consistent.

All verification above is Windows PowerShell 5.1.19041.6456. I haven't run a full four-version `deploy.ps1` yet — another IDE instance is holding the addin on this box — so the end-to-end deploy is still to come; everything above was exercised by extracting the real functions and running them under the real preference.

peterparker57 (approved):
Re-reviewed. All four points are addressed, and I verified the two that could only be settled by running something rather than reading it. Approving.

Where I could, I tested rather than took your word for it — not out of doubt, but because point 1 was a case where the code *reads* correct under either host and only a real 5.1 run tells you which one you have.

## Point 1 — verified, and the test discriminates

I extracted the real functions from both revisions via the AST and ran them under the `$ErrorActionPreference = "Stop"` that `deploy.ps1` sets on itself, on Windows PowerShell **5.1.26100.9168**, with `$sourceDir` pointing at an existing directory that is not a git tree — your `CLARIONLSP_ROOT`-at-an-unpacked-server-tree case.

Pre-fix (`eaaa2d2b`, `Assert-LspPin`):

```
git : fatal: not a git repository (or any of the parent directories): .git
    + FullyQualifiedErrorId : NativeCommandError
  >>> TERMINATED on the git line - soft-fail branch unreachable
```

Post-fix (PR head, `Test-LspPin` + `Invoke-GitQuiet`):

```
  WARN  ...\definitely-not-a-git-tree-9f3a is not a git tree — cannot verify it matches pin 571ffcee.
      returned: True
      REACHED-THE-LINE-AFTER
  >>> survived the git call (soft-fail branch ran)
```

Worth recording that my **first** run of that harness was worthless and I nearly reported it as a pass: I had `$ProjectDir` wrong by one path segment, so both arms died at `Get-Content` on the manifest before reaching the git line, and both printed "TERMINATED". Two identical failures look exactly like a confirmed bug when you want one. The A/B above is the corrected run, where the arms actually differ.

`Invoke-GitQuiet` itself is the right shape. Taking `[string[]]$GitArgs` rather than `ValueFromRemainingArguments` is the detail I'd have most wanted and least expected — a git flag can't be captured as a parameter name. The `finally` restoring the preference matters more than it looks, since the callers keep running under `Stop` afterwards and a leaked `Continue` would silently disarm every check downstream. And catching `CommandNotFoundException` separately is correct: git absent from PATH terminates regardless of the preference, so the `try/catch` isn't belt-and-braces there, it's the only thing covering that case.

Duplicating it in both scripts is the right call while `deploy.ps1` doesn't dot-source the sync. It is now a thing that can drift, so if a third consumer ever appears, that's the moment to extract it rather than paste it again.

## Point 2 — rebase confirmed independently

`1ceb6f7` is an ancestor of the PR head; merge base is `f3587fe`. And the reason I checked the second half too: master has moved **60 commits** since `f3587fe`, but **none of them touch `deploy.ps1` or `Sync-LspServer.ps1`**. So the rebase hasn't gone stale in the week this sat waiting on me — it still merges clean today, and the delay is mine, not a cost you have to pay again.

## Point 4 — all three taken

**`exit 7`.** `Test-LspPin` returning a bool is better than the `-SkipLspPin` switch I offered, and your reasoning for skip-the-copy over ship-anyway is right: a manifest asserting a commit the consumer didn't receive is precisely the drift the file exists to prevent, so shipping nothing is the honest failure. I traced `$LspPinOK` to its four consumers — the copy guard, the per-version report, and both halves of the summary — and the failure is named rather than folded into a generic exit. Renaming `Assert-` to `Test-` because it no longer throws is a small thing that will save someone a wrong assumption later.

**The fetch.** Now checked, with the offline chain written into the comment above it. That comment is worth more than the fix; the fix is three lines and obvious in hindsight, whereas "why this one unread exit status mattered when the others didn't" is the part that would have been re-litigated in two years.

**`Save-Manifest`.** Validating with `ConvertFrom-Json` before writing, and throwing rather than writing, is the fail-closed direction — the on-disk manifest stays good and the error names the writer instead of surfacing as a parse failure in `Read-Manifest` on the next run, pointing at the wrong file.

## One thing I checked that neither of us raised

Both files carry a UTF-8 BOM (87 and 51 non-ASCII bytes respectively), and both parse clean under real 5.1 — I checked with `Parser::ParseFile` under `powershell.exe`, not just by running them.

That is not idle. I have separately been bitten by a BOM-less UTF-8 `.ps1` in this repo with em dashes in it: 5.1 reads it as CP1252, `—` becomes `â€"`, and the trailing byte lands as U+201D — which PowerShell treats as a **string delimiter**, so a literal closes early and the parser reports a bogus missing-brace lines away from the real problem. Your files dodge it because they have the BOM. Since this PR adds prose with em dashes to both, please keep it that way — whatever editor you're using is doing the right thing, and it's worth not letting it change.

## Not blocking — pre-existing, and mine

While sweeping for leftovers I found the construct still live at `Sync-LspServer.ps1:321` and `:339` (`git remote get-url origin 2>$null`, and the `git show -s` inside the tag-exists message), plus the unchecked `git fetch` at `:330`. I confirmed all three are on master untouched by this PR — they're in the legacy `-Apply` overlay path, so they're my code, not yours, and they're reachable only from the re-pin path I run by hand.

They're the same bug class, so `:321` will terminate instead of printing the `Fail "origin is ..."` message on a clone with no `origin` remote. Please **don't** fold them in here — this PR is clean and I want it in before the release. I'll take them separately, and `Invoke-GitQuiet` is already sitting in the file waiting for them.

## Merge and release sequencing

Taking this now. It's release-relevant in the direction that matters: the commit check works immediately against the existing `resolvedCommit` (`571ffcee`), so merging this makes the imminent cut strictly better verified than it is today.

Then #191 — I've answered the reproducibility question there, and your hash matches mine byte for byte from a build made here on 2026-07-15, three weeks before you opened this. So rebase it and mark it ready. One sequencing caveat, which I've noted on that PR as well: #191 fails open on an existing manifest, so it verifies nothing until a `-Pure` run has recorded the hash. If the release lands between the merge and that run, it ships with the artifact unverified and a manifest that reads complete. That's on me to sequence, not a change request on you.

Thanks for this one. The diagnosis in the original PR was already the hard part, and the round since has been unusually easy to check — every claim came with the input that produces it.

bdinko:
Nudge on this one — it's carried an approval since 17 Aug, and I think it's now only waiting on the `changes requested` from 9 Aug to be cleared or re-reviewed. @ClarionLive, all four of your points were addressed back on the 10th; happy to do another round if anything still reads wrong.

Re-verified today rather than assuming: GitHub still reports `MERGEABLE / CLEAN`, and `git merge-tree` against current master (`784c9ed`) produces zero conflict markers.

**One correction to the record — conclusion unchanged.** The approval note said master had moved 60 commits "but none of them touch `deploy.ps1` or `Sync-LspServer.ps1`". One did: `6a737db` (*fix(deploy): deploy to ALL live installs of a Clarion version*), landed 13 Aug, four days before. It's an easy one to miss — filtering the log on `deploy.ps1` instead of `ClarionAssistant/deploy.ps1` returns zero hits and reads exactly like a clean result. I made that same mistake first and nearly reported it as a pass, which is a slightly on-the-nose echo of your own harness note.

It doesn't cost anything:

- **Textually disjoint.** `6a737db` works in `Resolve-ClarionRoot` → `Resolve-ClarionRoots` and the version/build loops; this PR works in `Resolve-LspBuild` and the LSP copy guard. Different regions of the same file, no overlap.
- **Semantically fine.** `$ResolvedRoots[$ver]` is an array now, and `$Roots` is iterated per install. `$LspPinOK` is computed once before the loops, so a pin failure still skips the copy for *every* install of every version — which is the behaviour the multi-root change wants, not something it breaks.

So still no rebase needed as far as I can see.

#191 is updated with the sequencing caveat you asked for — that merging it verifies nothing until the second `-Pure` run, and what that means for a release cut in between. It stays a draft until this one lands, then I'll rebase it and mark it ready, in the order you set out.

PR #191, opened 2026-08-10
fix(lsp): verify the shipped server.js, not just the source checkout (follow-up to #186 point 3)

bdinko:
Follow-up to your review point 3 on #186.

> Not a blocker — the check is still a strict improvement on nothing. But hashing the built `server.js` into the manifest is what would actually close the gap, and it'd be worth a follow-up.

**Draft, and stacked on #186.** It branches from `fix/lsp-pure-self-contained`, because the function it modifies (`Test-LspPin`) only exists there. Until #186 merges, the commit list below shows its four commits too — the payload here is the single commit `80b51f7`. Once #186 lands I'll rebase and mark this ready, and the duplicates disappear on their own.

> [!IMPORTANT]
> **Merging this verifies nothing on its own. The first green line it prints is not evidence.**
>
> `resolvedServerSha256` does not exist in today's `lsp-snapshot.json`, and this PR deliberately fails open on a manifest that lacks it. So the sequence is:
>
> | run | what happens | verified? |
> |---|---|---|
> | 1st `-Pure` after merge | **records** the hash into the manifest | **no** — warns and ships |
> | 2nd and after | hashes the shipped artifact and compares | yes |
>
> **Sequencing note for whoever cuts the next release.** A release cut between the merge and that second run ships with the artifact unverified *and* a manifest that now looks complete — the field is populated, so nothing in the output says the check never actually ran. If you are cutting that release, run `Sync-LspServer.ps1 -Pure` once first, commit the recorded hash, and let the release build be the run that is checked.
>
> This is a property of the fail-open design, not a defect in it — blocking a deploy over a field that didn't exist when the manifest was written would be worse. It is written here rather than only further down because the failure mode is someone reading a green line six months from now and trusting it.

## Why hashing alone isn't enough

You gave two cases. The hash closes the first and not the second, so this does both.

**Case 1 — an overlay build under a correct checkout.** `out/` is gitignored, survives `git checkout --force`, and `-Pure` skips the rebuild when a `server.js` already exists. So `CLARIONLSP_ROOT` pointing at an overlay clone at the pinned tag prints the green line while an overlay ships.

Fixed by hashing the artifact. `Sync-LspServer.ps1 -Pure` records `resolvedServerSha256` — sha256 of the exact `out/server/src/server.js` that `deploy.ps1` and `installer/ClarionAssistant.iss` copy. `Test-LspPin` hashes the file it is about to ship and compares; on a mismatch the LSP copy is skipped for every version, the same handling a commit mismatch now gets.

The commit check stays as a secondary signal. The hash says *this is the right file*; the commit says *and it came from the right tag*; a disagreement between the two is itself worth seeing, so both print.

**Case 2 — a re-pointed upstream tag.** Your words: *"refresh moves HEAD, the rebuild is skipped, resolvedCommit is rewritten, and the assert agrees with itself."* Hashing does **not** fix this. The sync would write the new commit **and** a hash of the old artifact — still self-consistent, still wrong.

So `-Pure` now stamps `out/.pure-build-stamp.json` with the commit that produced the build, and rebuilds when the stamp is missing or names a different commit. The stamp lives inside `out/` deliberately: "delete `out/` to force a rebuild" is the documented escape hatch, and this way a stale stamp can never outlive the build it describes.

## It fails open on an existing manifest

With no `resolvedServerSha256` the run warns that the shipped `server.js` is **not** verified and continues to the commit check, rather than blocking a deploy over a field that didn't exist when the manifest was written.

See the callout at the top for what that means for the first run after this merges, and for the release cut in between.

## Verification

Windows PowerShell 5.1.19041.6456. Real functions extracted from `deploy.ps1` via the AST and run under the file's own `$ErrorActionPreference = "Stop"`, against the real `.lsp-build/v1.0.0` tree:

| case | result |
|---|---|
| hash matches shipped artifact | `True` — green OK on both checks |
| hash mismatch (overlay / other-tag build) | `False` — LSP copy skipped |
| manifest without `resolvedServerSha256` | `True` — warns it is unverified |
| hash matches but source commit is wrong | `False` — commit check still bites |
| `CLARIONLSP_ROOT` override + hash mismatch | `True` — warns and ships |

End-to-end through the real `Sync-LspServer.ps1 -Pure -SkipBuild` against a non-git pure tree: writes both the stamp and `resolvedServerSha256`.

For the local `v1.0.0` build, `out/server/src/server.js` hashes to:

```
b16bf3db46f6b59ad6af6d1de438fd72136ff09019db5688fd5c6053becc8a30
```

(cross-checked against `Get-FileHash -Algorithm SHA256`.)

**Confirmed reproducible across machines** — @peterparker57 hashed his own `v1.0.0` tree and got the same 169818 bytes, byte-identical. That build was produced on 2026-07-15, three weeks before #186 was opened, and hashed where it sat rather than rebuilt to answer the question — so it is two independent `npm ci` + `npm run compile` runs, different machines, weeks apart, same digest.

That was the precondition for this design: a hash shared through the manifest is only safe if the build is byte-reproducible. It is, so the field can be a committed shared value rather than a per-machine one.

Still not committing a hash from my build — per the above, let the first `-Pure` run after the merge write it.

## Also in here

`Update-PinFields`' two git calls now go through `Invoke-GitQuiet`. On the supported non-git pure tree they printed two bare `fatal: not a git repository` lines into an otherwise clean run and recorded null anyway — same result, minus output that reads like a failure.

## Not addressed

`out/common` and the `node_modules` set are still unhashed — only `server.js` is, which is what you named. If you'd rather pin a manifest of every shipped file, that's a bigger change and I'd rather do it deliberately than smuggle it in here.

### DIFF
--- a/ClarionAssistant/deploy.ps1
+++ b/ClarionAssistant/deploy.ps1
@@ -277,32 +277,41 @@ function Invoke-GitQuiet([string[]]$GitArgs) {
     }
 }
 
-function Assert-LspPin($sourceDir) {
+# Returns $true when the LSP may ship, $false when it must not. It does NOT exit: an LSP-copy
+# problem is not a reason to abandon the Clarion builds. This used to `exit 7`, which aborted all
+# four Clarion targets AND the C# addin before a single one was built -- cutting straight across
+# the "a build failure for ONE Clarion release must not block shipping to the others" rule the
+# final summary below is built around. The caller skips just the LSP copy instead, so every version
+# still deploys and keeps whatever server it already had, and the run ends non-zero so nobody reads
+# it as clean.
+function Test-LspPin($sourceDir) {
     $manifest = Get-Content (Join-Path $ProjectDir "lsp-server-sync\lsp-snapshot.json") -Raw | ConvertFrom-Json
     $pinned   = $manifest.resolvedCommit
     if (-not $pinned) {
         Write-Host "  WARN  manifest has no resolvedCommit — cannot verify the bundled LSP." -ForegroundColor Yellow
-        return
+        return $true
     }
     $head = Invoke-GitQuiet @('-C', $sourceDir, 'rev-parse', '--short', 'HEAD')
     if (-not $head) {
         Write-Host "  WARN  $sourceDir is not a git tree — cannot verify it matches pin $pinned." -ForegroundColor Yellow
-        return
+        return $true
     }
     # Prefix compare: `git rev-parse --short` auto-scales its length, so 7- and 8-char forms of the
     # same commit must still count as a match.
     if (-not ($head.StartsWith($pinned) -or $pinned.StartsWith($head))) {
         if ($env:CLARIONLSP_ROOT) {
             Write-Host "  WARN  bundled LSP is $head but the pin says $pinned (CLARIONLSP_ROOT override in effect)." -ForegroundColor Yellow
-            return
+            return $true
         }
         Write-Host "  FAIL  bundled LSP is $head but lsp-snapshot.json pins $pinned." -ForegroundColor Red
         Write-Host "        Re-run lsp-server-sync\Sync-LspServer.ps1 -Pure, or bump the pin deliberately." -ForegroundColor Red
-        exit 7
+        Write-Host "        The LSP copy will be SKIPPED for every version; the rest of the deploy continues." -ForegroundColor Red
+        return $false
     }
     Write-Host "  OK    bundled LSP matches pin: $pinned ($($manifest.resolvedTag))" -ForegroundColor Green
+    return $true
 }
-Assert-LspPin $LspSourceDir
+$LspPinOK = Test-LspPin $LspSourceDir
 # Pure v1.0.0 runtime deps only — NO better-sqlite3/bindings/file-uri-to-path (those backed the retired
 # CodeGraph overlay). With better-sqlite3 absent the #42 ABI check below self-skips ("module not deployed").
 # iconv-lite (+ its dep safer-buffer) is NEW at v1.0.0: UnicodeDiagnostics requires it in the EAGER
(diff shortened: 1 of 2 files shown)


peterparker57:
Answering the question you flagged as blocking, since it decides whether the design survives.

**Your hash reproduces here.** My `v1.0.0` build:

```
b16bf3db46f6b59ad6af6d1de438fd72136ff09019db5688fd5c6053becc8a30
```

Byte-identical to yours.

Worth saying why that's a stronger signal than a bare match: this wasn't a build I ran to answer you. The tree at `.lsp-build/v1.0.0/out/server/src/server.js` was produced on this machine on 2026-07-15 — three weeks before you opened #186 — and I hashed it where it sat. So it's two independent `npm ci` + `npm run compile` runs, on different machines, weeks apart, landing on the same 169818 bytes. That's the condition a shared `resolvedServerSha256` needed, and you were right to make it a precondition rather than assume it.

You were also right not to commit a hash from your own build. Same answer now that it's confirmed: let the first `-Pure` run after the merge write it.

Which brings up the one thing I'd want stated in the PR body before this comes out of draft — you already say it, I just want it impossible to miss when someone reads the green line six months from now. **Merging this verifies nothing on its own.** An existing manifest has no `resolvedServerSha256`, so the first run warns-and-continues; only the run *after* the recording run is actually checked. If a release gets cut in between, it ships with the artifact unverified and a manifest that looks complete. Not a change request — a sequencing note for whoever cuts that release, which for the imminent one is me.

On the two design calls:

**The build stamp inside `out/`** is the right place. Tying the stamp's lifetime to the artifact it describes means "delete `out/` to force a rebuild" stays true and can't be silently subverted by a stamp that outlives its build. A stamp in the tree root would have been a second thing to keep in sync, i.e. a second thing to get wrong.

**Keeping both checks** is also right, and your reasoning is the reason: the hash says *this is the right file*, the commit says *and it came from the right tag*, and a disagreement between them is information neither check produces alone. My point 3 could have been read as "replace the commit check", and I'm glad it wasn't.

Not addressing `out/common` and the `node_modules` set here is the correct scope call. `server.js` is what I named; a full shipped-file manifest is a different change and deserves its own argument.

Nothing blocking from me. Get #186 in, rebase, and mark this ready — I'll re-review #186 now.

